PR #11, opened 2019-01-09
Switch to default Ant form label rendering

rkuykendall:


src/building-blocks/FormField.tsx
```diff
@@ -45,18 +42,11 @@ class FormField extends Component<IProps> {
...
-        </Label>
-        <Value>
-          <Antd.Form.Item>
-            {getFieldDecorator(fieldConfig.field, decoratorOptions)(
-              <EditComponent {...editProps} />,
-            )}
-          </Antd.Form.Item>
-        </Value>
-      </Info>
+      <Antd.Form.Item label={fieldConfig.label}>
```
kbouchard: I think you should keep the `key={}` , otherwise, 👍

kbouchard: fyi, I'm using this in the Registry:
`<Antd.Form.Item key={`field-config-${fieldConfig.field}-${index}`} label={fieldConfig.label}>`

PR #14, opened 2019-01-09
Bug fixes for ObjectSearchCreate component

rkuykendall:


src/inputs/ObjectSearchCreate.tsx
```diff
@@ -12,7 +12,6 @@ import { IFieldConfig, IFieldConfigObjectSearchCreate } from '../';
 interface IProps {
   fieldConfig: IFieldConfig;
   form: any;
-  isVisible?: any;
   label: string;
```
alexandermeyer: Why have a label prop if only the label in `fieldConfig` is used?

rkuykendall: This is a great question, I'm actually gonna look into this a bit more...

rkuykendall: Answer: There shouldn't be, label is never passed.

PR #18, opened 2019-01-10
Create first tests for Card, EditableCard component

rkuykendall:


src/components/EditableCard.tsx
```diff
@@ -3,7 +3,7 @@ import { observer } from 'mobx-react';
 import { observable } from 'mobx';
 import autoBindMethods from 'class-autobind-decorator';
 import SmartBool from '@mighty-justice/smart-bool';
-import { kebabCase } from 'lodash';
+import { kebabCase, noop } from 'lodash';
```
ronaldzuniga: Looks like not using `noop`

package.json
```diff
@@ -126,6 +131,7 @@
     "classnames": "^2.2.6",
     "date-fns": "^1.30.1",
     "flat": "^4.1.0",
+    "gitbook-plugin-search-plus": "^1.0.3",
```
ronaldzuniga: Why adding this?

rkuykendall: Ug, this is getting added automatically by some process. I haven't been able to figure it out yet. Maybe @kbouchard knows.

PR #33, opened 2019-01-14
Fix validation rules accumulating on calls of fillIn* functions

rkuykendall:


rkuykendall:
@kbouchard It's a very core change, but thankfully it's very easy to search our codebases for `formValidationRules` and neither ever use it. It's only used in the `TYPES` const right now. Plus, since the type changed, and we use Typescript, which is amazing and definitely speeds up and not slows down development, we will automatically be alerted if there are any outdated uses when we push out the library update.

PR #40, opened 2019-01-16
Add support for showLabel on forms

rkuykendall:


rkuykendall:
@ronaldzuniga Working on implementing storybook so we can include screenshots more easily. Those constraints on label also make a lot of sense to me. Less so the positioning, since that's more of a Keven thing.

PR #42, opened 2019-01-17
Add ChildrenBefore in FormModal

kbouchard:
This is something we have in v2, but not in fields-ant at the moment.

I need it now in the Registry and I'm sure we'll need it again in the future.


test/components/FormModal.test.js
```diff
@@ -1,6 +1,5 @@
 /* global it, describe, expect */
 
-import React from 'react';
```
kbouchard: 🤷‍♂️

Issue #36, opened 2019-01-15
Add radio button type


alexandermeyer:
As it stands now, we can use option select to achieve the same goals, but radio buttons are a field type we often use.

Adding this in fields-ant (along with using `insertIf`) would decrease the number of custom forms in our repositories

PR #44, opened 2019-01-18
Add option type and improve Storybook

rkuykendall:
Closes https://github.com/mighty-justice/fields-ant/issues/36

image

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -1,16 +1,69 @@
 import React, { Component } from 'react'
-import { configure, addDecorator } from '@storybook/react';
-import 'antd/dist/antd.css';
+import { Table, Tag } from 'antd';
+import { Provider } from 'mobx-react';
+import { pickBy } from 'lodash';
 
-const styles = {
-  margin: '15px',
-};
-const CenterDecorator = (storyFn) => (
-  <div style={styles}>
-    { storyFn() }
+import { configure, addDecorator } from '@storybook/react';
+import { withInfo } from "@storybook/addon-info";
+
+import 'antd/dist/antd.css';
+import './stories.css';
+
+const PropsTable = (props) => (
+  <Table
+    dataSource={props.propDefinitions}
+    pagination={false}
+    bordered
+    columns={[
+      {
+        dataIndex: 'property',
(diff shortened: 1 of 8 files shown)


.storybook/stories.css
```diff
@@ -0,0 +1,14 @@
+.story-component {
```
alexandermeyer: This looks cumbersome, did you write this style yourself? Is there a before after?

src/inputs/RadioGroup.tsx
```diff
@@ -0,0 +1,64 @@
...
+    if (this.injected.getOptions) { return this.injected.getOptions(optionType); }
+    return [];
+  }
+
+  private get options (): IOption[] {
+    return this.fieldConfig.sorted ? sortBy(this.unsortedOptions, 'name') : this.unsortedOptions;
+  }
+
+  public render () {
+    const radioStyle = {
```
alexandermeyer: You can define this outside of the class definition. It doesn't look like we have style files for our components in this repository yet, is that on the docket?

.storybook/config.js
```diff
@@ -1,16 +1,70 @@
...
+  ],
+};
+
+const ComponentDecorator = (storyFn) => (
+    <div className='story-component'>
+      { storyFn() }
+    </div>
+);
+
+addDecorator(ComponentDecorator);
```
alexandermeyer: Why do you need two decorators that both call `storyFn()`? I'm not familiar with storybook

rkuykendall: The decorators stack and wrap each other. So one needs to be before the withInfo, and one after.

.storybook/stories.css
rkuykendall: Added a comment. This adds the blueprint background behind components.

src/inputs/RadioGroup.tsx
rkuykendall: No, I think I'll just remove it here so it can be styled on-site like all our other components.

rkuykendall:
@alexandermeyer I considered breaking them up, but the PR is still very short. If you'd prefer though it could be split.

PR #53, opened 2019-01-24
Add populateFromSearch and populateNameFromSearch

rkuykendall:


src/building-blocks/NestedFieldSet.tsx
```diff
@@ -30,6 +35,47 @@ class NestedFieldSet extends Component<IProps> {
...
+        field,
+        populateFromSearch,
+        populateNameFromSearch,
+      } = fieldConfig;
+
+      if (populateFromSearch) {
+        defaults[field] = search;
+      }
+
+      if (populateNameFromSearch && field === 'first_name') {
```
alexandermeyer: Could these not be `===` but instead `endsWith`? I'm thinking of the case of, say, `plaintiff.first_name` and `plaintiff.last_name`.

The following line would have to change to something like `defaults[field] = firstName` (does `defaults` support the nested field name?)

src/interfaces.ts
```diff
@@ -19,6 +19,8 @@ interface IFieldConfigBase {
   key: string;
   label: string | null;
   nullify: boolean;
+  populateFromSearch: boolean;
```
alexandermeyer: Could there be an addition to the documentation about what these do? The names are relatively clear but could be confusing to the unacquainted

src/building-blocks/NestedFieldSet.tsx
rkuykendall: image

src/interfaces.ts
rkuykendall: Good point. Written.

docs/ObjectSearchCreate.md
```diff
@@ -0,0 +1,38 @@
...
+export interface IFieldConfigObjectSearchCreate extends IFieldConfigBase {
+  createFields: IFieldConfigBase[];
+  endpoint: string;
+  searchFilters?: { [key: string]: any };
+}
+```
+
+A FieldConfig of type `objectSearchCreate` defines:
+
+- `endpoint` which it will submit it's `?search=` queries to
```
alexandermeyer: `it's` --> `its`

docs/ObjectSearchCreate.md
```diff
@@ -0,0 +1,38 @@
...
+- `createFields` a FieldSet to allow the user to compose their own
+  object if they are unable to find one that matches their query
+- `searchFilters` (optional) params to include alongside the search
+  
+The create functionality is currently blocked until the user types  at least
```
alexandermeyer: Extra space before `at`

stories/types.stories.js
```diff
@@ -38,7 +40,33 @@ const props = {
   };
 
 storiesOf('Types', module)
-  .addDecorator(withInfoConfigured)
   .add('Displaying', () => <Card {...props} />)
   .add('Editing', () => <FormCard {...props} onSave={(data) => action('Form Save')(data)} />)
+  .add('objectSearchCreate', () => (
+    <>
+      <Marked md={'# { type: \'objectSearchCreate\' }'} />
```
alexandermeyer: You can use ` to prevent from escaping quotes?

docs/ObjectSearchCreate.md
```diff
@@ -0,0 +1,38 @@
...
+3 characters, but this will be customizable in the future.
+
+Create fields may include these optional attributes:
+
+- `populateFromSearch` to default the value of this field to the user's search 
+  string
+- `populateNameFromSearch` to default the value of a field ending in
```
alexandermeyer: This is a little unclear; it doesn't quite get at the idea that there can be both a `first_name` and a `last_name` marked with `populateNameFromSearch`. Though I guess with the example it indicates that

PR #54, opened 2019-01-24
Clean up objectSearchCreate tests

rkuykendall:
Other very small fixes:
- `IFieldSetPartial` is a super-type of `IFieldSetSimplePartial`, so this is an easy change
- `disabled` and `writeOnly` are filled in, therefore guaranteed in final `fieldConfig`

### DIFF
--- a/src/utilities/common.ts
+++ b/src/utilities/common.ts
@@ -89,6 +89,8 @@ export function fillInFieldConfig (fieldConfig: IFieldConfigPartial): IFieldConf
     } : undefined;
 
   return {
+    // Universal defaults
+    disabled: false,
     key: fieldConfig.field,
     label,
     readOnly: false,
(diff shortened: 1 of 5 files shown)


src/utilities/common.ts
```diff
@@ -89,23 +89,29 @@ export function fillInFieldConfig (fieldConfig: IFieldConfigPartial): IFieldConf
...
     ...TYPES[type],
     ...fieldConfig,
 
+    // Merge nested object
     editProps: {
       ...fieldConfig.editProps,
       ...TYPES[type].editProps,
     },
 
+    // Merge nested object
```
kbouchard: good job, the comments 🎖

test/types/objectSearchCreate.test.js
```diff
@@ -12,43 +12,65 @@ function changeInput (component, value) {
...
+    onSave,
+    props,
+    type,
+
+    ...overrides,
+  }
+}
+
+async function getTester (props) {
+  return (await new Tester(FormCard, { props }).mount());
```
kbouchard: I'm not sure how much that is saving but 🤷‍♂️

PR #56, opened 2019-01-24
Small improvements over almost entire test suite

rkuykendall:


test/types/optionSelect.test.js
```diff
@@ -10,27 +10,54 @@ const field = 'is_open'
...
       };
 
     const tester = await new Tester(FormCard, { props }).mount();
     expect(tester.text()).toContain(expectedLabel);
   });
+
+  it('Handles array with single value (useful for summary tables)', async () => {
+    const props = {
+        cardConfig: { fieldSets },
+        model: { is_open: ['true'] },
```
mumumumu: is this still necessary? since it's defined in the outer const
```suggestion
        model: { is_open: ['true'] },
```

rkuykendall: Good catch. the outterone should have been no-array.

PR #59, opened 2019-01-28
Allow to customize OSC Select & Button

kbouchard:
**note:** This might not be the best long term solution, but w/ Robert gone an given how complex that component is, I think this is the most simple and bug free solution for now.

We will revisit how we can properly allow customizing this component in the future.

osc-customizable

### DIFF
--- a/dist/inputs/ObjectSearchCreate.d.ts
+++ b/dist/inputs/ObjectSearchCreate.d.ts
@@ -6,6 +6,7 @@ declare class ObjectSearchCreate extends Component<IInputProps> {
     private search;
     private readonly injected;
     private readonly fieldConfig;
+    private readonly selectProps;
     private handleSearch;
     private addNew;
     private onChange;
--- a/src/inputs/ObjectSearchCreate.tsx
+++ b/src/inputs/ObjectSearchCreate.tsx
@@ -2,11 +2,14 @@ import React, { Component } from 'react';
 import { observable, toJS } from 'mobx';
 import { inject, observer } from 'mobx-react';
 import autoBindMethods from 'class-autobind-decorator';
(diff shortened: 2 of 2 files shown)


dist/inputs/ObjectSearchCreate.d.ts
```diff
@@ -6,6 +6,7 @@ declare class ObjectSearchCreate extends Component<IInputProps> {
     private search;
     private readonly injected;
     private readonly fieldConfig;
+    private readonly selectProps;
```
kbouchard: note: When the build command will run before deploying, the buttonProps will be added too:

screen shot 2019-01-28 at 10 07 56 am

src/inputs/ObjectSearchCreate.tsx
```diff
@@ -18,22 +21,38 @@ import {
 
 const MIN_SEARCH_LENGTH = 3;
 
+interface IProps {
+  fieldConfig: IFieldConfigObjectSearchCreate;
```
alexandermeyer: Order these props

src/inputs/ObjectSearchCreate.tsx
```diff
@@ -99,13 +119,13 @@ class ObjectSearchCreate extends Component<IInputProps> {
           ))}
         </Antd.Select>
         <Antd.Button
+          icon='plus'
```
alexandermeyer: Order these props (especially have `{...this.buttonProps}` come last)

kbouchard: I guess it's fine for now because I `pick()` them.
But normally, I wanted the ones at the top to be overridable and the ones at the bottom not to be.

PR #60, opened 2019-01-29
Add Rows and Columns logic in Forms

kbouchard:
It's likely that we review it when Robert gets back, but it feels like that's a clean enough solution for now.

Note: I wouldn't  worry about the coverage since the code "might" change.

screen shot 2019-01-29 at 9 33 22 am
screen shot 2019-01-29 at 9 33 26 am
screen shot 2019-01-29 at 9 33 57 am

### DIFF
--- a/stories/types.stories.js
+++ b/stories/types.stories.js
@@ -70,7 +70,7 @@ storiesOf('Types', module)
               },
               {
                 colProps: { sm: 24, lg: 12 },
-                field: 'example',
+                field: 'second_example',
                 type: 'objectSearchCreate',
                 endpoint: '/endpoint/',
                 createFields: [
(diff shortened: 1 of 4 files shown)


stories/types.stories.js
```diff
@@ -48,21 +48,44 @@ storiesOf('Types', module)
...
+                    type: 'objectSearchCreate',
+                    createFields: [{ field: 'name' }],
+                  },
+                ],
               },
-            ],
-          }]]
+              {
+                colProps: { sm: 24, lg: 12 },
+                field: 'example',
```
alexandermeyer: Could you change this field name to something like `example 2`, for clarity's sake? That way the two fields are different

PR #61, opened 2019-01-29
Improve FormDrawer customizability

kbouchard:
* Set specific `className` to help CSS customization.
* Remove hardcoded `width`, add it as a prop.

### DIFF
--- a/src/components/FormDrawer.tsx
+++ b/src/components/FormDrawer.tsx
@@ -22,9 +22,8 @@ interface IProps {
   onSave: (args: any) => Promise<void>;
   onSuccess?: (args?: any) => void;
   title: string;
+  width?: number;
 }
-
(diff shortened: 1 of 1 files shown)


src/components/FormDrawer.tsx
```diff
@@ -59,18 +58,19 @@ class BaseFormDrawer extends Component<IProps> {
   }
 
   public render () {
-    const { form, isVisible, model, title } = this.props;
+    const { form, isVisible, model, title, width } = this.props;
 
     return (
       <Drawer
+        className='mfa-form-drawer'
```
kbouchard: `mfa` --> `Mighty Fields Ant` 🤷‍♂️ 
Might change ? might stay ? Good for now ?

alexandermeyer: Good for now!

PR #66, opened 2019-02-04
Switch ObjectSearchCreate from using sub-form to existing form

rkuykendall:
Closes https://github.com/mighty-justice/fields-ant/issues/35

- ObjectSearchCreate - Fixes form submitting when sub-form invalid
- ObjectSearchCreate - Fixes required validation messages not showing
- FormDrawer - Fixes no required marks showing
- FormManager - Fixes form clearing on unsuccessful submit
- FormManager - Fixes onSubmit not waiting for validation to complete
- FormManager - Fixes incorrect fieldNames function not mapping errors to nested fields
- FormManager - Fixes toast notifications not rendering due to error object instead of string

image

image


src/building-blocks/NestedFieldSet.tsx
```diff
@@ -76,18 +76,19 @@ class NestedFieldSet extends Component<IProps> {
     return defaults;
   }
 
-  private async onFieldsChange (_props: any, fields: any) {
-    /*
-    This function is triggered on all fields changes
-    and includes values and errors of all sub-fields
```
rkuykendall: image

test/types/objectSearchCreate.test.js
```diff
@@ -94,9 +95,16 @@ describe('objectSearchCreate', () => {
     expect(tester.text()).toContain('required');
     expect(onSave).not.toHaveBeenCalled();
 
+    // Will not clear errors when changing valid field
+    changeInput(tester.find('input#amount_owed'), fakeOwed);
+    tester.find('form').simulate('submit');
+    expect(tester.text()).toContain('required');
+    expect(onSave).not.toHaveBeenCalled();
```
rkuykendall: Note: This test DID fail before the changes above were implemented.

rpalermodrums:
Looks good, but let's test this with `yalc` before merging/releasing

package.json
```diff
@@ -20,7 +20,6 @@
     "docs:deploy": "storybook-to-ghpages",
     "docs:watch": "start-storybook -p 6006",
     "lint": "tslint --project .",
-    "postpublish": "npm run docs:build && npm run docs:deploy",
```
rpalermodrums: reason? temporary for yalc or is this change permanent?

rkuykendall: yalc.

test/utilities/FormManager.test.js
```diff
@@ -0,0 +1,38 @@
...
+      'case.date_of_loss',
+      'case.state_of_incident',
+      'case.law_firm.name',
+      'plaintiff_treatment_status',
+    ];
+
+    const formManager = await getFormManager([fieldNames.map(field => ({ field }))]);
+    expect(formManager.formFieldNames).toEqual(fieldNames);
+  });
+});
```
rkuykendall: Note: This test DID fail before the changes above were implemented.

src/building-blocks/FormField.tsx
```diff
@@ -56,22 +60,27 @@ class FormField extends Component<IProps> {
...
       return null;
     }
 
-    const FormItemComponent = (
-      <Antd.Form.Item {...fieldConfig.formItemProps} label={this.label}>
-        {getFieldDecorator(fieldConfig.field, this.decoratorOptions)(
-          <fieldConfig.editComponent {...this.editProps} />,
-        )}
-      </Antd.Form.Item>
+    const decoratorOptionsProp = skipFieldDecorator ? { decoratorOptions: this.decoratorOptions } : {}
```
kbouchard: That whole block is quite complex ... too much ? I don't know.
I'm not sure how I'd clean it up either, so maybe it's fine as is.

src/building-blocks/NestedFieldSet.tsx
```diff
@@ -1,43 +1,44 @@
...
-    this.NestedForm = Antd.Form.create({
-      onFieldsChange: this.onFieldsChange,
-    })(FormFieldSet);
+    props.form.setFieldsValue({ [props.id]: {} });
   }
 
   @computed
   private get fieldSet () {
-    return getFieldSetFields(fillInFieldSet(this.props.fieldSet));
+    const { id } = this.props;
```
kbouchard: `, fieldSet`

src/building-blocks/NestedFieldSet.tsx
```diff
@@ -1,43 +1,44 @@
...
+    return getFieldSetFields(fillInFieldSet(this.props.fieldSet))
```
kbouchard: `return getFieldSetFields(fillInFieldSet(fieldSet))`

src/components/FormDrawer.tsx
```diff
@@ -72,12 +72,13 @@ class BaseFormDrawer extends Component<IProps> {
         visible={isVisible.isTrue}
         width={width}
       >
-        <Form layout='vertical' hideRequiredMark onSubmit={this.formManager.onSave}>
```
kbouchard: This is something we should discuss at some points.
I know we moved fast and that's fine, but the friction I often encounter with **fields-ant** is that I can't configure my components like how Ant Design is allowing me. Forcing me to modify fields-ant to change the hardcoded prop or allow me to pass in the Ant Design component's props.

It's an easy fix when we have 1 -> 1 components (`<ButtonXYZ />` for example). But when we have nested components like this, we can't simply allow one set of props. In that specific case we could allow `<FormDrawer>` to have extended `<Antd.Drawer />` props, but then what about the `<Antd.Form />` props ? We should try to find a way to allow it easily.

src/inputs/ObjectSearch.tsx
```diff
@@ -0,0 +1,96 @@
...
+    this.options = response.results;
+  }
+
+  private onChange (value: any) {
+    const foundOption = this.options.find(option => option.id === value.key);
+    this.injected.onChange(toJS(foundOption));
+  }
+
+  public render () {
+    const { id } = this.injected;
```
kbouchard: Why is the `id` injected ? 🤔

src/inputs/ObjectSearchCreate.tsx
```diff
@@ -96,37 +78,26 @@ class ObjectSearchCreate extends Component<IProps> {
...
-          disabled={this.search.length < MIN_SEARCH_LENGTH}
-          onClick={this.addNew}
-          {...this.buttonProps}
-        />
-      </Antd.Input.Group>
+      <Antd.Form.Item>
+        <Antd.Input.Group className='ant-input-group-search-create' compact>
+          {form.getFieldDecorator(this.props.fieldConfig.field, decoratorOptions)(
+            <ObjectSearch
+              fieldConfig={this.props.fieldConfig}
```
kbouchard: We might want to `const` all these `this.props.xyz`

src/utilities/FormManager.ts
```diff
@@ -108,26 +104,30 @@ class FormManager {
     this.notifyUserAboutErrors(errorMessages);
   }
 
-  public async onSave (event: any) {
-    const { form, onSave } = this.args;
+  private async validateThenSaveCallback (errors: any, _values: any) {
+      const { onSave } = this.args;
+      this.saving = true;
+      if (errors) { this.saving = false; return; }
```
kbouchard: Would `this.saving` profit from being a `SmartBool` or not really ? 🤔

src/building-blocks/NestedFieldSet.tsx
rkuykendall: ???

rkuykendall: OH, nevermind.

src/inputs/ObjectSearch.tsx
rkuykendall: Because this is an input, so this is injected by the input wrapper.

src/utilities/FormManager.ts
rkuykendall: I thought about it before, but the places it was being used were all too complex for a simple `.until(`. This seems to fit that use-case as well.

test/utilities/FormManager.test.js
```diff
@@ -0,0 +1,38 @@
+/* global describe, it, expect */
+/* eslint-disable no-magic-numbers, sort-keys */
```
alexandermeyer: We don't have eslint in this repository, right? Do you think we should add it (seeing `import {Tester}` below made me notice the lack of linting)?

rkuykendall: We do not! Must be a copy-paste problem. Also, I believe Jest now has native TS support so I'd like to attempt that and gain tslint, but registry blockers are obviously more important.

PR #77, opened 2019-02-07
Name and export component props

rkuykendall:
Here's the theory, from `props.ts`:
```
// Import and re-export component prop interfaces
// This cannot be re-exported from index.ts or rollup will fail
// These interfaces can be imported like so:
//
// import { IButtonToolbarProps } from '@mighty-justice/fields-ant/props';
```


rkuykendall:
@kbouchard I think just for distributed libraries... 🤔but I'm open to the idea. I'll show you how I plan to use these and maybe you can see if you would see if benefiting any other codebases...

PR #78, opened 2019-02-07
Add isLoading prop to FormCard

rkuykendall:
This brings `FormCard` in line with our other components:


src/components/FormCard.tsx
```diff
@@ -15,6 +15,7 @@ import { fillInFieldSets } from '../utilities/common';
 
 interface IExportProps extends ICommonCardProps {
   children?: any;
+  isLoading?: boolean;
```
alexandermeyer: Even though ordering isn't required, it's good form here!

PR #81, opened 2019-02-08
Improve coverage with more generic, factory-based testing

rkuykendall:
New / improved tests:
- `colProps.test.js`
- `insertIf.test.js`
- `rowProps.test.js`
- `types.test.js`
- `FormManager.test.js`
- `inferType.test.js`

Other small changes exposed by better testing:
- `CardField` BUGFIX: Should be able to accept partial `fieldConfig`
- `CardField` CLEANUP: Type-check for render method no longer necessary
- `CardFieldSet` CLEANUP: No need to pass filtered `fieldConfig`s to `CardField`
- `FormField` BUGFIX: Should be able to accept partial `fieldConfig`
- `ArrayCard` BUGFIX: Should render empty message when array empty, not just not render
- `inferType` CLEANUP: All fields with same name as a type should infer to that type
- `types` BUGFIX: Percentage must be cast to integer for validation


src/utilities/FormManager.ts
```diff
@@ -96,6 +96,7 @@ class FormManager {
   }
 
   private handleBackendResponse (response?: any) {
+    // istanbul ignore next
```
ronaldzuniga: Why did you have to add this? It's the only change in the file.

rkuykendall: It was an uncovered line that is not worth covering because it represents a 'This should never happen' type check / fallback error.

PR #82, opened 2019-02-11
Add all Components / Types to automated tests / docs

rkuykendall:
The main purpose if this PR is adding all the final types to `TYPE_GENERATORS` and all the final components to `COMPONENT_GENERATORS`. This allows all types and components to be included in some generic testing as well as added to the storybook documentation.

### DIFF
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -9,32 +9,50 @@ import { IForm, IValue } from './props';
 // much easier to call a function to fill in a partial definition
 // than to fill all this in (see further below for partials).
 interface IFieldConfigBase {
+  // Field is the ONLY required value of a fieldConfig
+  // All other attributes below are either optional or will be filled in
+  // by the component through the fillIn* functions
+  field: string;
+
+  // Another way of keeping fieldConfigs short are 'types' which are just
+  // different sets of defaults that can be quickly applied.
+  type: string;
+
+  // Core attributes
   className?: any;
-  colProps?: ColProps;
+  label: string | null;
+  nullify: boolean;
+  render: (...args: any[]) => string | JSX.Element | JSX.Element[];
+  showLabel: boolean;
+  value?: string | number;
+
+  // Attributes for filtering this field out, in read contexts,
+  // form contexts, or dynamically based on the model
+  insertIf?: (model: any) => boolean;
+  readOnly: boolean;
+  writeOnly: boolean;
+
+  // Attributes for controlling how fieldConfig works in form
   disabled: boolean;
   editComponent: any;
   editProps: { [key: string]: any };
-  field: string;
-  fieldConfigProp: boolean;
   formItemProps?: { [key: string]: any };
   formValidationRules: { [ruleName: string]: { [attribute: string]: any } };
   fromForm: (value: IValue) => any;
   icon?: string;
-  insertIf?: (model: any) => boolean;
-  key: string;
-  label: string | null;
-  nullify: boolean;
+  required: boolean;
+  toForm: (data: any, field: string) => any;
+
+  // These enable features in objectSearchCreate
+  // explained in the documentation for that type
   populateFromSearch: boolean;
   populateNameFromSearch: boolean;
-  readOnly: boolean;
-  render: (...args: any[]) => string | JSX.Element | JSX.Element[];
-  required: boolean;
-  showLabel: boolean;
+
+  // These are technical and you can safely ignore them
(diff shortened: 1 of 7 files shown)


src/interfaces.ts
```diff
@@ -9,32 +9,50 @@ import { IForm, IValue } from './props';
 // much easier to call a function to fill in a partial definition
 // than to fill all this in (see further below for partials).
 interface IFieldConfigBase {
+  // Field is the ONLY required value of a fieldConfig
```
kbouchard: 🙏

src/interfaces.ts
```diff
@@ -9,32 +9,50 @@ import { IForm, IValue } from './props';
...
+  // All other attributes below are either optional or will be filled in
+  // by the component through the fillIn* functions
+  field: string;
+
+  // Another way of keeping fieldConfigs short are 'types' which are just
+  // different sets of defaults that can be quickly applied.
```
kbouchard: 🙏

src/interfaces.ts
```diff
@@ -9,32 +9,50 @@ import { IForm, IValue } from './props';
...
+  type: string;
+
+  // Core attributes
   className?: any;
-  colProps?: ColProps;
+  label: string | null;
+  nullify: boolean;
+  render: (...args: any[]) => string | JSX.Element | JSX.Element[];
```
kbouchard: FYI, I noticed that Antd uses `React.ReactNode` and it works pretty well for anything if you don't need strict typing for children (which should be the case most of the time).
This should work: `render: (...args: any[]) => React.ReactNode;`
Except if I'm wrong :D

src/interfaces.ts
```diff
@@ -9,32 +9,50 @@ import { IForm, IValue } from './props';
...
-  insertIf?: (model: any) => boolean;
-  key: string;
-  label: string | null;
-  nullify: boolean;
+  required: boolean;
+  toForm: (data: any, field: string) => any;
+
+  // These enable features in objectSearchCreate
+  // explained in the documentation for that type
   populateFromSearch: boolean;
```
kbouchard: Question. If these are only for objectSearchCreate type, why would it be required ?
Would this be a more logical way to handle this ? `populateFromSearch?: boolean;`

Same thing for the next one or any type specific prop

rkuykendall:
```
// This excludes all fields used only by a single type ( see below )
// It also will almost never be written out by a user, since it's
// much easier to call a function to fill in a partial definition
// than to fill all this in (see further below for partials).

  // Field is the ONLY required value of a fieldConfig
  // All other attributes below are either optional or will be filled in
  // by the component through the fillIn* functions
```
If this explanation isn't enough ( which I don't think it is ), I'd like to tweak it until it is. However... I'm a little unsure what direction to go to improve it. Can you maybe try to elaborate?

src/interfaces.ts
rkuykendall: Nice find. I have a TS PR coming right after this so I'll add that to that one, as it seems better grouped there.

src/interfaces.ts
kbouchard: I read your comments above and I might be just more confused than I was 😆 
I don't want to block that PR or anything, so I'll try one last time to ask my question, in an other way.

Context: The way component props works is that we have required and optional props. To define it, we always do either `myProp: any;` OR `myProp?: any;` (`?` stating that it's optional).

Question: Based on what I said above, what benefit do we get from making it required, then setting it to `false` when filling it. 

Why the current :
```
{ populateFromSearch: boolean; }
...
fillIn() => {
  return {
  ...
    populateFromSearch: false
  }
}
```
Is better than the possible :
```
{ populateFromSearch?: boolean; }
...
fillIn() => {
  return {
  ...
  }
}
```
(No need to fill in in the second example since it's optional)

rkuykendall: Okay, I actually think you fully understand what's going on, which is good. So then we just have 2 options: `false | true` or `false | true | undefined`. I don't think in practice they should actually make any difference. We need to use the non-optional option for `showLabel` where the default is true. It seemed like we could do the same for false values and collapse our three-state attribute to a two-state attribute. However, I don't think it would make any difference removing these fills like you suggest.

kbouchard: [This conversation continued on Slack where it escalated really fast! We then started a meme fight to decide who was right and who was wrong. Robert won.]
- The End

PR #85, opened 2019-02-14
Add standard propDefaults for form components

rkuykendall:
Note that `propDefaults` make non-optional interface attributes optional for the caller.

So `onSubmit: blabla` + `onSubmit: asyncNoop` === `onSubmit?: blabla`


rkuykendall:
@ronaldzuniga This PR fixes a bug where `onSuccess` was required for FormModal where previously it wasn't. It fixes it in a way that also standardizes the propDefaults of all form components so that an oversight like that doesn't happen again.

rkuykendall:
@ronaldzuniga As seen when trying to upgrade fields-ant on our internal project:

image

ronaldzuniga:
Please link to the bug jira ticket in this case, thanks!

PR #86, opened 2019-02-14
Add form component wrapper components to fix defaultProps

rkuykendall:


src/components/FormModal.tsx
```diff
@@ -1,3 +1,4 @@
+// tslint:disable max-classes-per-file
```
rpalermodrums: ha, should we do this or actually make a separate file?

rkuykendall: 10 lines I say keep in-file.

It's not really a separate component so much as very complicated export to ensure props / default props make it to the output file correctly. Plus I think I can remove it with my next project, a basic Form component.

rpalermodrums: true, then should we get rid of this tslint rule?

rkuykendall: That's a really good question...

I'm leaning towards keeping it, because it makes us justify 'acceptable' use cases. But I'd be fine saying our tech team only does acceptable cases and killing it?

rpalermodrums: meh, i don't care enough to have an opinion really. I think that it's generally best to keep the rule and add an ignore for files that have a wrapper class as long as it never is expected to be the majority of the files

PR #87, opened 2019-02-15
Standardize and simplify forms with Form component

rkuykendall:
- Created standard `Form` component
- Changed components `FormCard`, `FormDrawer`, `FormModal` to use `Form`

## FormCard

- Form layout changed to vertical, removing label spacing and `:`
- Form controls aligned right
- Horizontal rule added over form controls

## FormDrawer

- Form controls aligned right
- Now properly implements `saveText`

## FormModal

- Form layout changed to vertical, removing label spacing and `:`

### DIFF
--- a/.storybook/config.tsx
+++ b/.storybook/config.tsx
@@ -8,33 +8,52 @@ import { withInfo } from '@storybook/addon-info';
 import 'antd/dist/antd.css';
 import './stories.css';
 
-const PropsTable = (props: any) => (
-  <Table
-    bordered
-    columns={[
-      {
-        dataIndex: 'property',
-        sorter: (a: any, b: any) => (a.property).localeCompare(b.property),
-        title: 'Prop',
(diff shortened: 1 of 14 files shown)


.storybook/config.tsx
```diff
@@ -8,33 +8,52 @@ import { withInfo } from '@storybook/addon-info';
...
+const PropsTable = (props: any) => {
+  const dataSource = props.propDefinitions.map((propDefinition: any) => ({
+    ...propDefinition,
+    required: propDefinition.required && !propDefinition.defaultValue,
+  }));
+
+  return (
+    <Table
+      bordered
+      columns={[
```
rpalermodrums: Nitpick, but I'd rather see this as a getter outside of the render function

rpalermodrums: I guess i really mean a `const`

Issue #67, opened 2019-02-05
insertIf does not show field on initial render


alexandermeyer:
I have a modal that uses `FormModal` with a `cardConfig` that uses `insertIf` for a field. This `insertIf` works great if the values in the modal are changed. On initial render, however, the model that is passed to the `insertIf` function is empty, even though fields have defaults.

rkuykendall:
I think this could be solved by moving initialValue logic into FormManager so that we can generate the initial model before the form is mounted.

PR #93, opened 2019-02-20
Move defaults calculation into FormManager

rkuykendall:
Part 2 of 3 on the road to https://github.com/mighty-justice/fields-ant/issues/67

### DIFF
--- a/src/utilities/FormManager.ts
+++ b/src/utilities/FormManager.ts
@@ -1,20 +1,34 @@
 import { observable } from 'mobx';
 import autoBindMethods from 'class-autobind-decorator';
-import { get, set, noop, mapValues, pickBy } from 'lodash';
-import flatten from 'flat';
+import flattenObject from 'flat';
+
+import {
+  flatten as flattenArray,
+  get,
+  mapValues,
+  noop,
+  pickBy,
+  set,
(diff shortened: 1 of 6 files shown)


src/utilities/FormManager.ts
```diff
@@ -82,7 +104,7 @@ class FormManager {
   }
 
   public get formFieldNames () {
-    return Object.keys(flatten<{ [key: string]: any }, { [key: string]: any }>(this.formValues));
+    return Object.keys(flattenObject<{ [key: string]: any }, { [key: string]: any }>(this.formValues));
```
kbouchard: Is this fixing a known bug or something ? Or it's just that ... it cause no issues but it should have been like this from the beginning ?

rkuykendall: The problem is that there are 2 libraries which both have 'flatten' as their suggested function name, and this file uses both. I could have just renamed one of them, but renaming both seemed the clearest option:
image

rkuykendall: So to clarify, not a bug-fix, not the way it always should have been (eh, maybe that one), but a differentiation because now we have 2 flatten functions in this file.

PR #99, opened 2019-02-21
Add loading state to ObjectSearchCreate

rkuykendall:


src/inputs/ObjectSearchCreateSearchInput.tsx
```diff
@@ -29,11 +32,21 @@ interface IEndpointOption {
   name: string;
 }
 
+/*
+This component performs the 'search' action of ObjectSearchCreate.
```
kbouchard: 👏

PR #102, opened 2019-02-22
Move ObjectSearchCreate Add New to dropdown

rkuykendall:
osc-v2

### DIFF
--- a/src/inputs/ObjectSearchCreateSearchInput.tsx
+++ b/src/inputs/ObjectSearchCreateSearchInput.tsx
@@ -176,7 +176,9 @@ class ObjectSearchCreateSearchInput extends Component<IObjectSearchProps> {
   }
 
   private onFocus () {
-    if (!this.hasOptions && !this.hasSearch) {
+    const isPristine = !this.hasOptions && !this.hasSearch;
+    if (isPristine) {
+      // Trigger empty search
       this.handleSearch(this.search);
     }
   }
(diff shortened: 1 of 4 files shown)


src/inputs/ObjectSearchCreateSearchInput.tsx
```diff
@@ -153,16 +165,26 @@ class ObjectSearchCreateSearchInput extends Component<IObjectSearchProps> {
       </Antd.Select.Option>
     );
   }
+  private onChange (selectedOption: any) {
```
alexandermeyer: Spacing before this function

src/inputs/ObjectSearchCreateSearchInput.tsx
```diff
@@ -153,16 +165,26 @@ class ObjectSearchCreateSearchInput extends Component<IObjectSearchProps> {
...
+    }
 
-  private onChange (value: any) {
-    const foundOption = this.options.find(option => option.id === value.key);
+    const foundOption = this.options.find(option => option.id === selectedOption.key);
     this.injected.onChange(toJS(foundOption));
   }
 
+  private onFocus () {
+    if (!this.hasOptions && !this.hasSearch) {
```
alexandermeyer: This `onFocus` is searching with an empty string to get a list of all possible options on initial click?

rkuykendall: I believe so. @kbouchard ?

kbouchard: ex-ac-tly

Focusing a "list" selection that would render no values at first was really confusing UX.

alexandermeyer: Then I would name a boolean something like that so it's clearer to the next person who reads it!

PR #106, opened 2019-02-26
Enable IFieldConfigBase.className and apply it on <Antd.Form.Item>

kbouchard:
## JIRA
https://thatsmighty.atlassian.net/browse/REG-197

## Screenshots
screen shot 2019-02-26 at 3 28 43 pm
screen shot 2019-02-26 at 3 28 53 pm

### DIFF
--- a/src/inputs/ObjectSearchCreate.tsx
+++ b/src/inputs/ObjectSearchCreate.tsx
@@ -24,6 +24,7 @@ import ObjectSearchCreateSearchInput from './ObjectSearchCreateSearchInput';
 
 export interface IObjectSearchCreateProps {
   addNewContent?: React.ReactNode;
+  className?: string;
   debounceWait?: number;
   decoratorOptions: { [key: string]: any };
   fieldConfig: IFieldConfigObjectSearchCreate;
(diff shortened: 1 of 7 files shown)


src/inputs/ObjectSearchCreate.tsx
```diff
@@ -115,6 +110,24 @@ class ObjectSearchCreate extends Component<IObjectSearchCreateProps> {
...
+
+  public render () {
+    const className = cx(
+      CX_PREFIX_SEARCH_CREATE,
+      {[`${CX_PREFIX_SEARCH_CREATE}-create`]: this.isAddingNew.isTrue },
+      this.props.className,
+    );
+
+    return (
+      <div className={className}>
```
kbouchard: Now properly setting `className` for both search and add modes, on the same level, same component.

src/inputs/Rate.tsx
```diff
@@ -4,14 +4,24 @@ import { EMPTY_FIELD } from '@mighty-justice/utils';
 
 import * as Antd from 'antd';
```
kbouchard: I've replicate how the RadioGroup is structured because `class Rate extends Component<any>` scared me 😱 .

But I might be wrong ...

kbouchard:
@rkuykendall  Oh ! you're 100% right. I was initially doing that on a new prop, but then decided to use the existing `className` and forgot to change it. Basically, we should use what `className`s are and have always been, `string`s.

If you look at Antd typing, all `className`s are `string`. So we should do that too.

screen shot 2019-02-26 at 3 50 36 pm

src/inputs/ObjectSearchCreate.tsx
```diff
@@ -76,36 +77,30 @@ class ObjectSearchCreate extends Component<IObjectSearchCreateProps> {
...
+  private renderAddNew () {
+    const { fieldConfig, formManager } = this.injected;
+
+    return (
+      <>
+        <NestedFieldSet
+          fieldSet={this.fieldConfig.createFields}
+          formManager={formManager}
+          id={fieldConfig.field}
+          label={this.fieldConfig.label}
```
rkuykendall: This seems messy. No pulling things out of fieldConfigs, and no consistency of `this.fieldConfig` vs { fieldConfig } = this.props;` above. I'd be fine with us pulling all out of this.fieldConfig, or just using this.fieldConfig all over, but mixing seems bad.

rkuykendall: Actually I just realized this is a whitespace diff so ignore 😛

kbouchard: @rkuykendall Well, I didn't change any of the props, I simply split the Add and the Search rendering pieces for easier reading of the `render()` function. + to have a proper main `className` on the root element of that component.

PR #107, opened 2019-02-26
Add types: url, ssn, phone, password, hidden

rkuykendall:
# Displaying
image

# Editing
image
image
image
image

### DIFF
--- a/src/utilities/types.ts
+++ b/src/utilities/types.ts
@@ -30,6 +30,7 @@ import OptionSelect from '../inputs/OptionSelect';
 import RadioGroup from '../inputs/RadioGroup';
 import Rate, { formatRating } from '../inputs/Rate';
 import { formatOptionSelect } from '../inputs/OptionSelectDisplay';
+import { REGEXP_PHONE, REGEXP_SSN } from '../consts';
 
 function stripFieldConfig (func: (...args: any[]) => any) {
   // tslint:disable-next-line no-unnecessary-callback-wrapper
@@ -154,7 +155,7 @@ export const TYPES: { [key: string]: Partial<IFieldConfig> } = {
     formValidationRules: {
       isPhoneNumber: {
         message: 'Must be a valid phone number',
-        pattern: /^\(?([0-9]{3})\)?[-. ]?([0-9]{3})[-. ]?([0-9]{4})$/,
+        pattern: REGEXP_PHONE,
         type: 'regexp',
       },
     },
@@ -176,7 +177,7 @@ export const TYPES: { [key: string]: Partial<IFieldConfig> } = {
     formValidationRules: {
       ssn: {
         message: 'Must be a valid social security number',
-        pattern: /^[0-9]{3}[-\s]?[0-9]{2}[-\s]?[0-9]{4}$/,
+        pattern: REGEXP_SSN,
         type: 'regexp',
       },
     },
--- a/src/utilities/types.ts
+++ b/src/utilities/types.ts
@@ -10,10 +10,14 @@ import {
 } from '../interfaces';
 
 import {
+  EMPTY_FIELD,
   formatCommaSeparatedNumber,
   formatDate,
   formatMoney,
   formatPercentage,
+  formatPhoneNumber,
+  formatSocialSecurityNumber,
+  formatWebsite,
   getNameOrDefault,
   getPercentDisplay,
   getPercentValue,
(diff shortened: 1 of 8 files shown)


test/features/insertIf.test.tsx
```diff
@@ -139,7 +139,10 @@ describe('insertIf', () => {
 
   it('Recieves data from model', async () => {
```
alexandermeyer: I just noticed: all these `Receives` are misspelled. May as well fix when you're adjusting these tests

test/factories.ts
```diff
@@ -120,6 +121,26 @@ export const optionSelectFactory = new Factory()
     type: 'optionSelect',
   });
 
+export const hiddenFactory = new Factory()
```
alexandermeyer: How about creating a `typeFactory` function that takes a type argument, since this most of these have no other attributes? Not necessary but it would clean this section up considerably

test/types/types.test.tsx
```diff
@@ -38,6 +40,10 @@ Object.keys(TYPE_GENERATORS).forEach(type => {
 
       // Renders empty placeholder
       const withoutData = await new Tester(Card, { props: { ...props, model: {} }}).mount();
+      if (fillInFieldConfig(fieldConfig).writeOnly) {
+        expect(withoutData.text()).not.toContain('--');
```
alexandermeyer: This should reference `EMPTY_FIELD`

test/factories.ts
rkuykendall: I hate that this never occurred to me. That cleanup was INCREDIBLE.

src/utilities/types.ts
```diff
@@ -133,6 +149,17 @@ export const TYPES: { [key: string]: Partial<IFieldConfig> } = {
     render: stripFieldConfig(formatPercentage),
     toForm: (data: any, field: string) => getPercentDisplay(get(data, field)),
   },
+  phone: {
+    editComponent: Antd.Input,
+    formValidationRules: {
+      isPhoneNumber: {
+        message: 'Must be a valid phone number',
+        pattern: /^\(?([0-9]{3})\)?[-. ]?([0-9]{3})[-. ]?([0-9]{4})$/,
```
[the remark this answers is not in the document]
rkuykendall: I kinda like that you can see the entire field definition, including validation, in one place? But I'm fine pulling regex out. Maybe into consts?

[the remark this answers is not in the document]
alexandermeyer: I agree with the entire field definition being there, including the validation rules. The only thing I suggest moving is the regex patterns here and below since regex is often 1. reused 2. tweaked slightly. Putting them in consts sounds good for now

src/utilities/types.ts
```diff
@@ -152,4 +191,15 @@ export const TYPES: { [key: string]: Partial<IFieldConfig> } = {
...
+  url: {
+    editComponent: Antd.Input,
+    editProps: { type: 'url' },
+    formValidationRules: {
+      url: {
+        message: 'Not a valid website (URLs should start with http:// or https://)',
+        type: 'url',
+      },
+    },
+    render: stripFieldConfig(formatWebsite),
```
alexandermeyer: A rename of this function should probably be on our radar. I assume it was originally meant to do just that - strip an optional field config parameter - but now it's being used to strip all sorts of optional parameters and I have to look up what it does every time I review 😆

test/factories.ts
```diff
@@ -206,15 +175,20 @@ export const TYPE_GENERATORS: ITypeGenerators = {
   date: { valueFunction: fakeDateRecent, fieldConfigFactory: dateFactory },
```
alexandermeyer: The next step for this testing improvement would be to have a fill in function for `TYPE_GENERATORS` that would automatically set the `fieldConfigFactory` to `fieldFactoryForType(FIELD_NAME)` so factory declarations wouldn't be needed except in special cases

src/utilities/types.ts
rkuykendall: Na, it's being used to strip the second param here, the fieldConfig param. What name would be more clear?

test/factories.ts
rkuykendall: Worth considering.

src/utilities/types.ts
alexandermeyer: Something like `stripOptionalParameters` or `stripArguments`, as this function does not deal with fieldConfigs specifically

PR #108, opened 2019-02-26
ObjectSearchCreate: Back to search should reset field value

rkuykendall:


src/inputs/ObjectSearchCreate.tsx
```diff
@@ -89,7 +91,7 @@ class ObjectSearchCreate extends Component<IObjectSearchCreateProps> {
           label={this.fieldConfig.label}
           search={this.search}
         />
-        <Antd.Button size='small' onClick={this.onSearch}>
+        <Antd.Button className='btn-back' size='small' onClick={this.onSearch}>
```
kbouchard: The way Ant Design structure their classNames is always by prepending the component's className , then the sub-component one. I started doing this to reflect as much as possible how Ant Design is structure to avoid confusion.

In that case, it would be :
`${CX_PREFIX_SEARCH_CREATE}-create`

We might want to look into how we could improve how we apply these classNames (avoid repetition), but ... I don'T think it's a big deal for now.

PR #110, opened 2019-02-27
Allow OSC to render children

kbouchard:
## JIRA
https://thatsmighty.atlassian.net/browse/REG-196
Note: At the beginning I wanted that for any fields-ant fields, but I don't think it make much sense for things like <Input />s, <RadioButton />s.

The goal is for OSC to display extra data when an item is selected, which is more likely to happen with OSC vs anything else.

screen shot 2019-02-27 at 10 19 19 am

### DIFF
--- a/src/building-blocks/ButtonToolbar.tsx
+++ b/src/building-blocks/ButtonToolbar.tsx
@@ -6,7 +6,6 @@ import { Form } from 'antd';
 
 export interface IButtonToolbarProps {
   align?: 'between' | 'right';
-  children?: React.ReactNode;
   className?: any;
   fixed?: boolean;
   noSpacing?: boolean;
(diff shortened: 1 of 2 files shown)


src/building-blocks/ButtonToolbar.tsx
```diff
@@ -6,7 +6,6 @@ import { Form } from 'antd';
 
 export interface IButtonToolbarProps {
   align?: 'between' | 'right';
-  children?: any;
```
kbouchard: This was useless. No need to define it when extending Component except if you want to force it or force specific children type.

src/inputs/ObjectSearchCreate.tsx
```diff
@@ -128,6 +128,7 @@ class ObjectSearchCreate extends Component<IObjectSearchCreateProps> {
             ? this.renderAddNew()
             : this.renderSearch()
         }
+        {this.props.children}
```
kbouchard: This was hard ...

PR #111, opened 2019-02-27
Infer phone_number as phone, not number

rkuykendall:


src/utilities/common.ts
```diff
@@ -62,6 +62,10 @@ function inferType (fieldConfig: Partial<IFieldConfig>) {
   if (field.endsWith('_at')) { return 'date'; }
   if (field.startsWith('is_')) { return 'boolean'; }
 
+  // Putting this ahead of loop so phone_number => phone, and not number
+  if (field.includes('phone')) { return 'phone'; }
```
kbouchard: Maybe one day 'phone' || 'fax' || 'cell' would be useful too, but no big deal.

PR #113, opened 2019-02-27
Implement readOnly attribute; fix button spacing

rkuykendall:


src/components/Form.tsx
```diff
@@ -67,7 +67,7 @@ export class UnwrappedForm extends Component<IFormWrappedProps> {
       <>
         <Antd.Divider />
 
-        <ButtonToolbar align='right'>
+        <ButtonToolbar align='right' noSpacing>
```
ronaldzuniga: So passing this prop to the component makes it use the new `.button-toolbar.no-spacing` class?

rkuykendall: Yeah, not so much new, I've just never used it, but it seems commonly used in our app.

PR #117, opened 2019-03-07
Add Table component

rkuykendall:
- New `Table` component that acts like `ArrayCard`
- if `onCancel` isn't passed, don't show a cancel button
- Remove divider between buttons and end of form

# In storybook
image

# Form 
image


src/components/Form.tsx
```diff
@@ -43,7 +42,7 @@ export class UnwrappedForm extends Component<IFormWrappedProps> {
         defaults,
         model,
         onSave,
-        onSuccess: onCancel,
+        onSuccess: this.onSuccess,
```
kbouchard: I guess it's fine that `this.onSuccess()` returns undefined if both are not set ?

src/components/FormDrawer.tsx
```diff
@@ -15,6 +15,7 @@ import Form from './Form';
 
 export interface IFormDrawerProps extends ISharedComponentProps, ISharedFormProps {
   isVisible: SmartBool;
+  onCancel: () => void;
```
kbouchard: `onCancel?: () => void;` shouldn't that be optional ?

.storybook/stories.css
```diff
@@ -18,7 +18,8 @@
   box-shadow: 0px 3px 10px rgba(0, 0, 0, 0.5);
 }
 
-#story-root > .ant-form {
+#story-root > .ant-form,
+#story-root > .ant-table-simple {
```
kbouchard: `ant-table-simple` is a lie. It doesn't exist in AntD.
It's a custom version I build for the registry.

You can simply use `.ant-table`

src/components/Form.tsx
rkuykendall: Yeah, onCancel and onSuccess are both void functions. this.onSuccess doesn't return anything no matter what.

src/components/Table.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  private title () {
+    return this.props.title;
+  }
+
+  public render () {
+    return (
+      <Antd.Table
+        {...this.props}
+        className='ant-table-simple'
```
kbouchard: This shouldn't be here, it's registry related. Also, to me `fields-ant` should not define styling in any way.
(My comment above says that you can just target the regular `.ant-table`, does the same job.)

FYI, we should always think about how this can be used. for example, the way it's done now, I'd have no way to set a className or a custom title because it would be overwritten because of the props order.

This would be more adequate:
```
<Antd.Table
    title={this.title}
    {...this.props}
    columns={this.columns}
    dataSource={this.dataSource}
/>
```

src/global.d.ts
```diff
@@ -1,5 +1,3 @@
-declare module '@mighty-justice/tester';
```
kbouchard: 🎉

src/components/FormDrawer.tsx
rkuykendall: For every component but FormDrawer, which will ALWAYS show a [X] at the top.

src/components/Table.tsx
rkuykendall: Title has to come after this.props since title is a string but Table wants a function for title.

PR #118, opened 2019-03-08
Fix title always showing on Table

rkuykendall:


src/components/Table.tsx
```diff
@@ -29,17 +30,25 @@ class Table extends Component<ITableProps> {
...
+
+  private propsWithoutTitle () {
+    return omit(this.props, 'title');
   }
 
   public render () {
     return (
       <Antd.Table
-        {...this.props}
+        {...this.propsWithoutTitle}
```
kbouchard: That's a lot just to handle one little prop. Could we reduce that by a getter that would return `tableProps` and only inject the title one when it's defined ?

+ You're setting functions, but not triggering them ? 🤔 
`this.propsWithoutTitle -> fn()`
`this.propsWithoutTitle() -> {...props}`

```ts
private get tableProps () {
  return {...omit(this.props, 'title'), title: this.props.title ? this.getTitle : null};
}

...
<Antd.Table
  {...this.tableProps}
  columns={this.columns}
  dataSource={this.dataSource}
/>
```
Wouldn't that work ?

PR #119, opened 2019-03-11
Add insertIf feature to table columns

rkuykendall:


src/utilities/common.ts
```diff
@@ -184,9 +184,12 @@ export function renderValue (fieldConfigPartial: IFieldConfigPartial, model?: IM
   return render(fieldConfig.value || get(model, field), fieldConfig, model || {});
 }
 
-export function fieldSetsToColumns (fieldSets: IFieldSetPartial[]): Array<ColumnProps<IModel>> {
+type IColumns = Array<ColumnProps<IModel>>;
+
+export function fieldSetsToColumns (fieldSets: IFieldSetPartial[], tableModel: IModel[] = []): IColumns {
   return getFieldSetsFields(fillInFieldSets(fieldSets))
     .filter(fieldConfig => !fieldConfig.writeOnly)
+    .filter(fieldConfig => !filterInsertIf(fieldConfig, tableModel))
```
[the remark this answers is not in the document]
rkuykendall: Totally agree on tightening up the nature of data. I don't love the fact that `model: IModel | IModel[]` but it seems like the easiest way to maintain a consistent-ish API at this time.

test/features/insertIf.test.tsx
```diff
@@ -137,6 +137,26 @@ describe('insertIf', () => {
     expect(none.html()).not.toBe(null);
   });
 
+  it('Works with individual CardFields', async () => {
```
SHA-4: You can instantiate one Tester with three different fields in its fieldset, then check if those field labels are present, rather than instantiating three testers

rkuykendall: 👍

test/features/insertIf.test.tsx
```diff
@@ -138,23 +139,22 @@ describe('insertIf', () => {
...
-        model: [],
-      }}).mount()
-      , none = await new Tester(Table, { props: {
-        fieldSets: [[{ field, label: exampleLabel }]],
+    const labelHidden = fakeTextShort()
+      , labelShowing = fakeTextShort()
+      , labelNone = fakeTextShort()
+      , tester = await new Tester(Table, { props: {
+        fieldSets: [[
+          { type: 'string', field: `hidden`, label: labelHidden, insertIf: jest.fn(_values => false) },
```
SHA-4: Field names can be with ' rather than `

test/features/insertIf.test.tsx
```diff
@@ -138,23 +139,22 @@ describe('insertIf', () => {
...
+          { type: 'string', field: `none`, label: labelNone },
+        ]],
         model: [],
       }}).mount()
       ;
 
-    expect(hidden.text()).not.toContain(exampleLabel);
-    expect(showing.text()).toContain(exampleLabel);
-    expect(none.text()).toContain(exampleLabel);
+    expect(tester.text()).not.toContain(labelHidden);
```
SHA-4: This `tester.text()` can be drawn out into a `const`, but not necessary

PR #125, opened 2019-03-18
Add support for isLoading on Table

rkuykendall:


src/components/Table.tsx
```diff
@@ -35,12 +35,17 @@ class Table extends Component<ITableProps> {
   }
 
   public render () {
+    const { isLoading, title } = this.props;
+
+    console.log('render.isLoading', isLoading);
```
wpontius: Leftover?

rkuykendall: Should have done self-CR -_-

PR #128, opened 2019-03-18
Respect falsey value attributes on render

rkuykendall:


test/factories.ts
```diff
@@ -196,6 +196,27 @@ export const TYPE_GENERATORS: ITypeGenerators = {
...
+const SKIP = null;
+
+export const valueRenderPairs: { [key: string]: [IValue, string | null] } = {
+  // value: [valueFunction(), valueFunction()]
+  ...fromPairs(Object.keys(TYPE_GENERATORS).map(type => {
+    const value = TYPE_GENERATORS[type].valueFunction();
+    return [type, [value, value]];
+  })),
+
+  boolean: sample([[true, 'Yes'], [false, 'No']]) as [boolean, string],
```
ronaldzuniga: Looks like we should be using faker instead of hard-coded strings

rkuykendall: The purpose of this dict is actually to hardcode strings in cases where `value !=== render`. The purpose is to test all the render functions for each type. There's no way to do it with faker without calling the render functions at which point we would no longer be testing the render functions.

Issue #98, opened 2019-02-21
ObjectSearchCreate not nullifying createFields


rkuykendall:

ronaldzuniga:
Can you post a video or steps to reproduce and expected values to understand what's exactly the issue?

rkuykendall:
@ronaldzuniga I'll write a failing test now. Though technically I think there's already lots of tests in the codebase that aren't failing but should be.

rkuykendall:
@ronaldzuniga Added failing test to https://github.com/mighty-justice/fields-ant/pull/134/files

PR #130, opened 2019-03-20
Move model processing into utility

rkuykendall:
Preliminary PR on the way to https://github.com/mighty-justice/fields-ant/issues/98


rkuykendall:
@ronaldzuniga Description includes link to ticket that this gets closer to fixing. `fields-ant` issues are tracked in public on Github.

PR #131, opened 2019-03-20
Bump all libraries and upgrade to Storybook 5

rkuykendall:


.storybook/webpack.config.js
```diff
@@ -3,12 +3,15 @@
 // after https://github.com/webpack/webpack/issues/3460 will be resolved.
 const TSDocgenPlugin = require('react-docgen-typescript-webpack-plugin');
 
-module.exports = (baseConfig, env, defaultConfig) => {
-  defaultConfig.module.rules.push({
+module.exports = ({ config, mode }) => {
```
SHA-4: Why have mode here if it's unused?

rkuykendall: Just standards? Make it easier to use mode in the future? I copied Keven's upgrade of this same file on Registry. I'm fine removing it.

PR #134, opened 2019-03-21
Nullify appropriate createFields in objectSearchCreate

rkuykendall:
Closes https://github.com/mighty-justice/fields-ant/issues/98

### DIFF
--- a/src/utilities/common.ts
+++ b/src/utilities/common.ts
@@ -214,8 +214,11 @@ export function fieldSetsToColumns (fieldSets: IFieldSetPartial[], tableModel: I
 
 export function modelFromFieldConfigs (fieldConfigs: IFieldConfigPartial[], data: IModel) {
     /*
-    This function takes in a model with ALL form values, including
-     */
+    This function takes in a model with ALL form values, including those that should be hidden like
+    readOnly fieldConfigs and those hidden by insertIf. We build a new model from scratch, only
+    including those that should be there. We also nullify falsey values that require it here, and
+    include the id from the model even if there is no fieldConfig for it.
+    */
     const returnValues: IModel = {};
 
     fieldConfigs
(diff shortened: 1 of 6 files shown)


src/utilities/common.ts
```diff
@@ -213,11 +223,16 @@ export function modelFromFieldConfigs (fieldConfigs: IFieldConfigPartial[], data
         const { field, nullify } = fieldConfig
           , formValue = get(data, field)
           , shouldNullify = nullify && !formValue && formValue !== false
-          , value = shouldNullify ? null : formValue
-          ;
+          , value = isTypeObjectSearchCreate(fieldConfig) && isObject(formValue) && !has(formValue, ID_ATTR)
```
SHA-4: Could you split this logic into another boolean so the name can provide some context for this? For example, I don't understand why `!has(formValue, ID_ATTR)` is important on whether to recurse

src/utilities/common.ts
```diff
@@ -213,11 +223,16 @@ export function modelFromFieldConfigs (fieldConfigs: IFieldConfigPartial[], data
...
+            ? modelFromFieldConfigs(fieldConfig.createFields, formValue)
+            : shouldNullify ? null : formValue
+            ;
 
         set(returnValues, field, value);
       });
 
+    const id = get(data, ID_ATTR);
```
SHA-4: Why are these lines being added? If the `id` was to be in the model, wouldn't it be a field?

rkuykendall: It's funny, yeah, it would seem that way. But if we stopped including ids from model unless they had a ( hidden? ) field defined, most of CoOp and registry would stop working. This is something we've done since the original V2 FormManager. Most patch functions patch to `/${data.id}` without ever thinking about how they never included an id in their field sets. It's just one of those things you always expect to be there.

src/utilities/common.ts
rkuykendall: I tried my best. I agree it's kinda awkward. Let me know if this helps.

PR #137, opened 2019-04-02
Add processErrors prop to control error messages on Forms

rkuykendall:


src/utilities/FormManager.ts
```diff
@@ -173,7 +175,7 @@ class FormManager {
 
   private notifyUserAboutErrors (errors: IErrorMessage[]) {
     errors.forEach(({ field, message }) => {
-      const description = `${field} - ${message}`;
+      const description = [field, message].filter(s => !!s).join(' - ');
```
rpalermodrums: 👍

PR #143, opened 2019-04-05
Show option search by default if scrollable

rkuykendall:
JIRA: https://thatsmighty.atlassian.net/browse/MTY-537

### DIFF
--- a/src/inputs/OptionSelect.tsx
+++ b/src/inputs/OptionSelect.tsx
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import { computed } from 'mobx';
 import { inject } from 'mobx-react';
+import { isBoolean } from 'lodash';
 
 import * as Antd from 'antd';
 
@@ -35,6 +36,18 @@ class OptionSelect extends Component<IOptionSelectProps> {
     return getOptions(this.fieldConfig, this.injected);
   }
 
+  private get showSearch (): boolean {
+    // 8 is the most number of options you can show with no scroll
+    const SHOW_SEARCH_IF_OVER = 8;
+
+    // the showSearch fieldConfig option will override this
+    if (isBoolean(this.fieldConfig.showSearch)) {
+      return this.fieldConfig.showSearch;
+    }
+
+    return this.options.length > SHOW_SEARCH_IF_OVER;
+  }
+
   public render () {
     const { renderOption } = this.props;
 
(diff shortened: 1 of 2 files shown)


src/inputs/OptionSelect.tsx
```diff
@@ -35,14 +36,26 @@ class OptionSelect extends Component<IOptionSelectProps> {
     return getOptions(this.fieldConfig, this.injected);
   }
 
+  private get showSearch (): boolean {
+    // 8 is the most number of options you can show with no scroll
+    const SHOW_SEARCH_IF_OVER = 8;
```
SHA-4: Do we normally define constants like these in functions? I thought we put them at the top of the file generally, even if they are only used once in the component

test/types/optionSelect.test.tsx
```diff
@@ -58,4 +61,32 @@ describe('optionSelect', () => {
...
+
+    function getPropsFor (overrides: Partial<IFieldConfig> = {}) {
+      return { props: { fieldSets: [[{ ...fieldConfig, ...overrides }]] } };
+    }
+
+    function isSearchable (tester: Tester) {
+      return !!tester.find('input').length;
+    }
+
+    async function expectIsSearchable (overrides: Partial<IFieldConfig>, tobe: boolean)  {
```
SHA-4: I think `tobe` would be clearer as `expectedValue`. If you disagree heavily, then you can keep it but please camel case

test/types/optionSelect.test.tsx
```diff
@@ -58,4 +61,32 @@ describe('optionSelect', () => {
     expect(tester.text()).not.toContain('Yes');
     expect(tester.text()).not.toContain('2 Values');
   });
+
+  it('Shows search for large data sets, or on command', async () => {
+    // tslint:disable no-magic-numbers
+    const states = Array(50).map((_v: any) => ({
```
SHA-4: This `50` should be replaced by something based off `SHOW_SEARCH_IF_OVER` like `SHOW_SEARCH_IF_OVER + 1`

test/types/optionSelect.test.tsx
```diff
@@ -58,4 +61,32 @@ describe('optionSelect', () => {
...
+        name: fakeTextShort(),
+        value: faker.random.uuid(),
+      }))
+      , fieldConfig = { field, options: states, optionType: 'state', type };
+
+    function getPropsFor (overrides: Partial<IFieldConfig> = {}) {
```
SHA-4: Why not `getProps`, `getOverriddenProps`, `getPropsWithOverrides`? I find prepositions at the end of function names usually point to 1. an unneeded word or 2. a missing noun

PR #145, opened 2019-04-09
Update dependencies based on learnings from mobx-react package.json

rkuykendall:
- Mobx should be 4 or 5
- peerDependencies should include .0.0
- Dependencies for things the parent might not have
- Various library bumps


package.json
```diff
@@ -117,52 +110,52 @@
...
     "rosie": "^2.0.1",
     "source-map-explorer": "^1.8.0",
     "storybook-readme": "^5.0.1",
     "style-loader": "^0.23.1",
-    "tslint": "^5.14.0",
-    "typescript": "^3.3.4000"
+    "tslint": "5.14.0",
+    "typescript": "3.3.4000"
   },
   "dependencies": {
```
SHA-4: How are we separating `devDependencies` and `dependencies`? The difference feels arbitrary here (but maybe I'm missing something)

package.json
```diff
@@ -117,52 +110,52 @@
...
-    "moment": "^2",
-    "react": "^16"
+    "@mighty-justice/smart-bool": "1.0.0",
+    "@mighty-justice/utils": "1.2.1",
+    "class-autobind-decorator": "3.0.1",
+    "classnames": "2.2.6",
+    "date-fns": "1.30.1",
+    "flat": "4.1.0",
+    "http-status-codes": "1.3.2",
+    "iso8601-duration": "1.2.0"
```
SHA-4: Why the restrictions for these packages? I understand for packages like `tslint` and `typescript`

package.json
rkuykendall: - `dependency` - When this library is installed, Install dependencies in a 'black box' way for use by this library. It's unlikely the parent library has this installed, and if they don't, I don't care, this one is mine.
- `peerDependency` - The parent should already have this library, with these specifications.
- `devDependency` - This library is either a `peerDependency` and should be installed when working on this library on it's own, or it's only used for building or documentation or something.

package.json
rkuykendall: Good point

PR #146, opened 2019-04-12
Sizing and validation tweaks for Birthdate

rkuykendall:
Better at small sizes, as low at 480px:

### DIFF
--- a/src/inputs/Birthdate.tsx
+++ b/src/inputs/Birthdate.tsx
@@ -30,13 +30,13 @@ const inputConfig: {
   year: IInputConfig,
 } = {
   day: {
-    style: { width: '27%', marginRight: '3%' },
+    style: { width: '29%', marginRight: '3%' },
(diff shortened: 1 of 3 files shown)


src/inputs/Birthdate.tsx
```diff
@@ -30,19 +30,19 @@ const inputConfig: {
   year: IInputConfig,
 } = {
   day: {
-    style: { width: '23%', marginRight: '3%' },
+    style: { width: '29%', marginRight: '3%' },
```
SHA-4: 29% seems like an arbitrary number... why not a round 30%?

rkuykendall: `29-29-36` (7) is almost twice as wide a difference in month/day and year as `30-30-34` (4). So it's true, 30 is a round number, but a single percentage point makes a big difference in this case.

PR #151, opened 2019-04-19
Reduce use of fillIn methods to speed up forms

rkuykendall:


src/utilities/FormManager.ts
```diff
@@ -78,9 +78,8 @@ class FormManager {
     return getFieldSetsFields(this.args.fieldSets);
   }
 
-  public getDefaultValue (fieldConfigPartial: IFieldConfigPartial) {
+  public getDefaultValue (fieldConfig: IFieldConfig) {
```
SHA-4: What has changed to ensure that these are `fieldConfig`s, not `fieldConfigPartial`s? Have these always been full `fieldConfig`s?

stories/types.stories.tsx
```diff
@@ -24,6 +25,34 @@ const props = {
...
+      )}
+    />
+  ))
+  .add('Form Stress Test (text)', () => (
+    <FormCard
+      {...formCardPropsFactory.build()}
+      {...omit(props, 'fieldSets')}
+      fieldSets={[
+        // tslint:disable-next-line no-magic-numbers
+        Array(159)
```
SHA-4: I realize where this is from, but we should probably have it be a more round number for it to be less magical

stories/types.stories.tsx
```diff
@@ -24,6 +25,34 @@ const props = {
   };
 
 storiesOf('Types', module)
```
SHA-4: How should I change my reviewing approach when looking at a storybook test versus a standard test? Is it just a render test? Could you explain better what I'm looking at?

src/utilities/FormManager.ts
rkuykendall: It looks like it was never actually called with a Partial. I was just leaving it open to make it as usable as possible but it was wasting cycles filling in non-partials all the time.

stories/types.stories.tsx
rkuykendall: I'd say much lower standards, since it's _not_ a test. It's like... documentation? This one here is just a development helper. It makes it easier to do things locally.

stories/types.stories.tsx
rkuykendall: Hmmm... I could raise it to 160 but I think 200 would be too high and would make it no longer representative. Since this is our largest known use-case It's not that magical? I could lower it to 150 but then it kinda loses it's usefulness.

SHA-4: I'd say make it 175, especially because our use-case is dynamic

PR #162, opened 2019-04-30
Improve renderOption, renderSelected attributes in ObjectSearchCreate

rkuykendall:
**JIRA:** https://thatsmighty.atlassian.net/browse/MTY-963
Closes https://github.com/mighty-justice/fields-ant/issues/156

image
image


src/interfaces.ts
```diff
@@ -166,5 +168,4 @@ export interface IInputProps {
 
 export interface IEndpointOption {
   id: string;
-  name: string;
```
SHA-4: Why fully remove name here? It looks like it's used in `getNameOrDefault`

rkuykendall: It is but since that can be overridden it MIGHT NOT HAVE NAME :o It could have... first_name? or something?

Issue #159, opened 2019-04-29
Support complex field sets in createFields for ObjectSearchCreate


SHA-4:
Right now, `rowProps` and `legend` are stripped from `createFields` in `NestedFieldSet.tsx`.

PR #163, opened 2019-05-01
Allow complex nested field sets

SHA-4:
Allows for `rowProps` and `legend` to be passed into `FormFieldSet`.

Fixes https://github.com/mighty-justice/fields-ant/issues/159.

### DIFF
--- a/src/utilities/common.ts
+++ b/src/utilities/common.ts
@@ -165,14 +165,7 @@ export function mapFieldSetFields (fieldSet: IFieldSetPartial, mapper: IMapper):
 
 export function fillInFieldSet (fieldSet: IFieldSetPartial): IFieldSet {
   // Fills in the defaults from common so we can keep configurations light
-  if (isPartialFieldSetSimple(fieldSet)) {
-    return fieldSet.map(fillInFieldConfig);
-  }
-
-  return {
-    ...fieldSet,
-    fields: fieldSet.fields.map(fillInFieldConfig),
-  };
+  return mapFieldSetFields(fieldSet, fillInFieldConfig) as IFieldSet;
 }
 
 export function fillInFieldSets (fieldSets: IFieldSetPartial[]): IFieldSet[] {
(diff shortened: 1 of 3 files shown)


src/utilities/common.ts
```diff
@@ -150,6 +150,18 @@ export function fillInFieldConfig (fieldConfig: IFieldConfigPartial): IFieldConf
   };
 }
 
+export function mapFieldSetFields (fieldSet: IFieldSetPartial,
```
SHA-4: I was thinking that `fillInFieldSet` is actually a `mapFieldSetFields` with `fillInFieldConfig`, but I didn't want to change something so integral without buy-in

rkuykendall: DO EEET

rkuykendall: image

src/utilities/common.ts
```diff
@@ -150,6 +150,18 @@ export function fillInFieldConfig (fieldConfig: IFieldConfigPartial): IFieldConf
...
+                                   mapper: (fields: IFieldConfigPartial) => IFieldConfigPartial): IFieldSetPartial {
```
rkuykendall: I hate this indentation style. I hate it on the backend too, but that's their problem.

rkuykendall: Another idea: 'setFieldSetFields` which would mirror `getFieldSetFields` and then `mapFieldSetFields` could be like... `setFieldSetFields(fieldSet, getFieldSetFields(fieldSet).map(mapper))` or is that dumb?

SHA-4: I don't think it's a bad idea, but wouldn't want to implement it right away for this feature, if that makes sense. First let's see the need and go from there?

rkuykendall: ¯\_(ツ)_/¯

rkuykendall: image

Issue #160, opened 2019-04-29
Support popovers in legends and labels


SHA-4:
We only support strings when we label fields and field sets, which means we cannot add popovers / tooltips. Popover content should be able to be passed as well as popover children

PR #173, opened 2019-05-02
Support popovers in complex FieldSet legends and fieldConfigs

rkuykendall:
**JIRA:** https://thatsmighty.atlassian.net/browse/MTY-985
Closes: https://github.com/mighty-justice/fields-ant/issues/160

image
image


src/building-blocks/CardFieldSet.tsx
```diff
@@ -28,24 +29,42 @@ class CardFieldSet extends Component<ICardFieldSetProps> {
...
+    const fieldSet = this.fieldSet;
+
+    if (isFieldSetSimple(fieldSet)) { return null; }
+    const { legend, tooltip } = fieldSet;
+
+    if (!legend) { return null; }
+
+    return (
+      <h3>
+        {tooltip
```
ronaldzuniga: Just wondering why this is not more like react nested components, like instead of having a function called renderWithToolTip not having something like:
```javascript
return (
   { shouldRenderTooltip ? <ToolTip> <Legend /> </ToolTip> : <Legend /> }
);
```

src/building-blocks/FormFieldSet.tsx
```diff
@@ -41,9 +42,26 @@ class FormFieldSet extends Component<IFormFieldSetProps> {
     return fieldConfigs.filter(fieldConfig => !filterInsertIf(fieldConfig, formModel));
   }
 
+  private renderLegend () {
```
ronaldzuniga: Duplicate code, please refactor.

src/utilities/renderWithTooltip.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+import { Icon, Tooltip } from 'antd';
+
+export function renderWithTooltip (label: string | null, tooltip: string) {
```
ronaldzuniga: Could easily be a `<ToolTip />` component that takes a prop to show tooltip or not.

PR #174, opened 2019-05-02
Add objectSearch type

rkuykendall:
**JIRA:** https://thatsmighty.atlassian.net/browse/MTY-987

image
image

### DIFF
--- a/test/factories.ts
+++ b/test/factories.ts
@@ -95,8 +95,7 @@ export const objectSearchCreateFactory = fieldFactoryForType('objectSearchCreate
   .extend(objectSearchFactory)
   .attrs({
     createFields: [{ field: 'name', required: true }],
-    endpoint: '/endpoint/',
-  });
+    });
 
 /*
  *   FIELD SET FACTORY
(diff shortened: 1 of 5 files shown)


test/factories.ts
```diff
@@ -86,7 +86,13 @@ export const radioFactory = fieldFactoryForType('radio')
 export const optionSelectFactory = fieldFactoryForType('optionSelect')
   .attrs({ options: attrOptions });
 
+export const objectSearchFactory = fieldFactoryForType('objectSearch')
+  .attrs({
+    endpoint: '/endpoint/',
+  });
+
 export const objectSearchCreateFactory = fieldFactoryForType('objectSearchCreate')
+  .extend(objectSearchFactory)
```
SHA-4: Wouldn't this extension override the type of `objectSearchCreateFactory`?

test/factories.ts
```diff
@@ -86,7 +86,13 @@ export const radioFactory = fieldFactoryForType('radio')
...
   .attrs({
     createFields: [{ field: 'name', required: true }],
     endpoint: '/endpoint/',
```
SHA-4: If you do end up extending `objectSearchFactory`, this attribute is redundant

test/factories.ts
rkuykendall: I suppose so?

test/factories.ts
rkuykendall: I suppose so?

test/factories.ts
rkuykendall: Seems fine

Issue #172, opened 2019-05-02
When clicking create in OSC, value is not being changed


rkuykendall:
See `new_case_contact.gif` in Robert's downloads folder ( not public gif )

PR #178, opened 2019-05-06
Bugfix: When clicking create in OSC, value is not being changed

rkuykendall:
**JIRA:** https://thatsmighty.atlassian.net/browse/MTY-1006
Closes https://github.com/mighty-justice/fields-ant/issues/172

### DIFF
--- a/test/types/objectSearchCreate.test.tsx
+++ b/test/types/objectSearchCreate.test.tsx
@@ -7,12 +7,15 @@ import { fakeTextShort } from '../factories';
 
 import { objectSearchFor } from './objectSearch.test';
 
-function getDefaults (overrides?: any) {
-  const fakeLawFirm = () => ({
-      id: faker.random.uuid(),
+function fakeLawFirm () {
+  return {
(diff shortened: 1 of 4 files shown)


test/types/objectSearchCreate.test.tsx
```diff
@@ -103,6 +106,32 @@ describe('objectSearchCreate', () => {
...
+    expect(tester.text()).toContain('Required');
+
+    // Will clear errors when fixing invalid field
+    tester.changeInput('input[id="law_firm.name"]', searchTerm);
+    tester.submit();
+    expect(onSave).toHaveBeenCalledWith({ law_firm: { name: searchTerm, amount_owed: null }});
+
+    tester.click('.ant-input-search-create-btn-back');
+    tester.submit();
+    expect(onSave).toHaveBeenCalledWith(props.model);
```
SHA-4: It seems unintuitive for this to be filled upon clicking back. What's more intuitive is that each state change (to add new, to search create) zeroes out the submit data

Issue #177, opened 2019-05-06
Clearing object search create input does not clear render on first click


SHA-4:
The model is cleared on first click, but the selected option still shows until the second click

PR #179, opened 2019-05-07
Bugfix: Clearing object search create input does not clear render on first click

rkuykendall:
**Jira:** https://thatsmighty.atlassian.net/browse/MTY-1009
Closes https://github.com/mighty-justice/fields-ant/issues/177

### DIFF
--- a/test/types/objectSearch.test.tsx
+++ b/test/types/objectSearch.test.tsx
@@ -2,16 +2,52 @@ import faker from 'faker';
 
 import { Tester } from '@mighty-justice/tester';
 
-import { fillInFieldConfig } from '../../src';
+import { fillInFieldConfig, FormCard } from '../../src';
 import ObjectSearch from '../../src/inputs/ObjectSearch';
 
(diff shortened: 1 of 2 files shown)


test/types/objectSearch.test.tsx
```diff
@@ -2,16 +2,52 @@ import faker from 'faker';
...
+    result: fakeLawFirm(),
+    searchTerm: faker.lorem.sentence(),
+
+    endpoint,
+    field,
+    model,
+    props,
+    type,
+
+    ...overrides,
```
ronaldzuniga: shouldn't this go at the end after `fieldConfig`? Also, not sure about the whitespaces here, but don't look good.

rkuykendall: `overrides.fieldConfig` is already included in the fieldConfig, which is then filled in. So this removes all the defaults from fieldConfig and makes it not filled in... if that makes any sense.

PR #182, opened 2019-05-08
Clean up fields-ant utilities

rkuykendall:


src/utilities/fillIn.ts
```diff
@@ -0,0 +1,118 @@
+import { IFieldConfig, IFieldConfigPartial, IFieldSet, IFieldSetPartial } from '../interfaces';
```
SHA-4: Tier your imports!

rkuykendall: It'll be in the follow-up!

PR #184, opened 2019-05-09
Add readOnly and writeOnly to filterFieldConfig

rkuykendall:


src/building-blocks/FormFieldSet.tsx
```diff
@@ -39,7 +39,8 @@ class FormFieldSet extends Component<IFormFieldSetProps> {
     }
 
     const formModel = this.props.formManager.formModel;
-    return fieldConfigs.filter(fieldConfig => !filterFieldConfig(fieldConfig, { model: formModel }));
+    return fieldConfigs.filter(
+      fieldConfig => !filterFieldConfig(fieldConfig, { model: formModel, readOnly: true }));
```
SHA-4: Was this supposed to have a `readOnly` before / why is this logic being added now?

src/building-blocks/CardFieldSet.tsx
```diff
@@ -33,7 +33,8 @@ class CardFieldSet extends Component<ICardFieldSetProps> {
     const { model , fieldSet} = this.props
       , idx = this.props.idx || 0
       , fieldConfigs = getFieldSetFields(this.fieldSet)
-      , filteredFieldConfigs = fieldConfigs.filter(fieldConfig => !filterFieldConfig(fieldConfig, { model }))
+      , filteredFieldConfigs = fieldConfigs.filter(
+        fieldConfig => !filterFieldConfig(fieldConfig, { model, writeOnly: true }))
```
SHA-4: Same question about why this logic is being added

src/building-blocks/FormFieldSet.tsx
rkuykendall: Yes, I believe this was an oversight. The filtered-out fieldConfigs would never be rendered due to being caught by the lower-level components, but in very rare cases like a FieldSet with 1 read-only fieldConfig, this would have caused an empty fieldset to render.

I'm gonna write a test for it I think.

src/building-blocks/CardFieldSet.tsx
rkuykendall: Answered above

src/building-blocks/FormFieldSet.tsx
rkuykendall: Yes this was an oversight. No, I can't test it yet, but will be able to in next PR.

PR #185, opened 2019-05-09
Move ArrayCard create logic into Form component

rkuykendall:


ronaldzuniga:
Why is this change required? Is it fixing a bug or part of a bigger feature? Can you please put link to the issue this is fixing? Thanks!

PR #187, opened 2019-05-10
Add formModel prop to Form sub-components for significant speed increases

rkuykendall:
The changes made for https://thatsmighty.atlassian.net/browse/MTY-1035 totally destroyed the speed increases we made to support 150+ input forms. This PR brings performance back in line with before the change.

`formManager.formModel` is by far the most expensive operation when rendering. This change significantly reduces the number of times it has to be called, down to near 1 per render.

### DIFF
--- a/src/building-blocks/FormFieldSet.tsx
+++ b/src/building-blocks/FormFieldSet.tsx
@@ -14,6 +14,7 @@ import {
 } from '../utilities';
 
 import { IFieldSetPartial } from '../interfaces';
+import { IModel } from '../props';
 
 import FormField from './FormField';
 import Legend from './Legend';
@@ -21,6 +22,7 @@ import Legend from './Legend';
 export interface IFormFieldSetProps {
   fieldSet: IFieldSetPartial;
   formManager: FormManager;
+  formModel: IModel;
 }
 
 @autoBindMethods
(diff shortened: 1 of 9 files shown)


src/building-blocks/FormFieldSet.tsx
```diff
@@ -31,22 +33,10 @@ class FormFieldSet extends Component<IFormFieldSetProps> {
...
-    }
-
-    // One of them does, so get formModel and filter out insertIfs
-    const formModel = this.props.formManager.formModel;
-    return filterFieldConfigs(fieldConfigs, { model: formModel, readOnly: true });
-  }
-
   public render () {
-    const filteredFieldConfigs = this.filteredFieldConfigs
+    const { formModel } = this.props
```
rpalermodrums: Could all of this be done inside of the constructor or is it necessary to be computed the time of render? Doing this once at the beginning of the lifecycle would likely also have significant performance improvements if it's possible

src/components/Form.tsx
```diff
@@ -110,20 +106,23 @@ export class UnwrappedForm extends Component<IFormWrappedProps> {
   }
 
   public render () {
-    const { showControls, title } = this.props;
+    const { showControls, title } = this.props
+      , formModel = this.formManager.formModel
+      , filteredFieldSets = filterFieldSets(this.fieldSets, { model: formModel });
```
rpalermodrums: basically same comment as above

src/building-blocks/FormFieldSet.tsx
rkuykendall: Yeah but insertIf takes the form model as an input and can ( and usually does ) return different results when the form model changes.

src/components/Form.tsx
rkuykendall: Leaving conversation above.

PR #197, opened 2019-05-20
Standardize props / behavior between FormModal and FormDrawer

rkuykendall:
**Jira:** https://thatsmighty.atlassian.net/browse/MTY-1094
**Jira:** https://thatsmighty.atlassian.net/browse/MTY-1095

- Create `ISharedFormModalProps` for both components to use
- Add `isVisible` SmartBool` to `FormModal`
- Make `isVisible` optional in both modals, default to always showing
- Automatically call isVisible.setFalse on `onSuccess` and `onCancel` if they are not specified
- Default side-modal width to 600px, to minimize the number of required props needed.

### DIFF
--- a/test/components/FormDrawer.test.tsx
+++ b/test/components/FormDrawer.test.tsx
@@ -1,23 +1,5 @@
-import faker from 'faker';
-
-import { FormDrawer } from '../../src';
-import { Tester } from '@mighty-justice/tester';
-import SmartBool from '@mighty-justice/smart-bool';
-
 describe('FormDrawer', () => {
-  it('Renders', async () => {
-    const text = faker.lorem.sentence()
-      , title = faker.lorem.sentence()
-      , props = {
-        fieldSets: [[{ field: 'text' }]],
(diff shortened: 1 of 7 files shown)


test/components/FormDrawer.test.tsx
```diff
@@ -1,23 +1,5 @@
...
-        isVisible: new SmartBool(true),
-        model: { text },
-        title,
-      };
-
-    const tester = await new Tester(FormDrawer, { props }).mount();
-    expect(tester.text()).toContain(title);
-    expect(tester.text()).toContain('Text');
-    expect(tester.html()).toContain(text);
+  it('is tested the same as FormModal', async () => {
```
rpalermodrums: Shouldn't this file be removed?

rkuykendall: I thought about it, but someone looking to test this component may not know where to look and I couldn't think of a better think to rename `FormModal.test.tsx` to indicate it tested both FormModal and FormDrawer.

PR #199, opened 2019-05-20
Add ModalComponent prop to EditableCard

rkuykendall:
**Jira:** https://thatsmighty.atlassian.net/browse/MTY-1096


src/components/EditableCard.tsx
```diff
@@ -9,12 +9,15 @@ import SmartBool from '@mighty-justice/smart-bool';
 import ButtonToolbar from '../building-blocks/ButtonToolbar';
 import GuardedButton from '../building-blocks/GuardedButton';
 import { formPropsDefaults } from '../propsDefaults';
-import { ISharedFormProps } from '../props';
+import { ISharedFormModalProps, ISharedFormProps } from '../props';
```
rpalermodrums: Not really relevant to this PR, but we can probably remove `Shared` from these titles at this point

PR #202, opened 2019-05-24
Bugfix: Hidden fields no longer render anything on the page

rkuykendall:
**Jira:** https://thatsmighty.atlassian.net/browse/MTY-1137

## Before
Screenshot 2019-05-24 11 17 20

## After
Screenshot 2019-05-24 11 14 34


src/inputs/Hidden.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+import {
+  IAntFormField,
+  IInjected,
+  IInputProps,
+} from '../interfaces';
+
+@autoBindMethods
+@observer
+class Hidden extends Component<IInputProps> {
```
ronaldzuniga: `HiddenInput`? Just `Hidden` sounds too generic, in fact, after reading the name I expected a component to hide any children based on a condition or something. But then noticed that his is actually an input.

rkuykendall: We do Rate and Date without input 🤔

src/utilities/types.ts
```diff
@@ -113,10 +114,11 @@ export const TYPES: { [key: string]: Partial<IFieldConfig> } = {
     },
   },
   hidden: {
-    editComponent: Antd.Input,
-    editProps: { type: 'hidden' },
+    editComponent: Hidden,
+    fieldConfigProp: true,
```
karen-ka: How did this change help resolve the bug? Was it because hiding the edit component still left the space? And what do the newly added properties do?

src/inputs/Hidden.tsx
ronaldzuniga: I think we should rename them too... 😊 What are they? Date Picker? Date range? Date only? Date and time? Rate gauge? Rate like in 5 stars? Is it an input or a control? But for the scope of this particular PR Hidden is just too ambiguous.

rkuykendall: All of these are inputs. They are in the `inputs/` folder, but if you think we should siffix all of them with `*Input.tsx` I could understand that:

image

src/utilities/types.ts
rkuykendall: @karen-ka That is a great question. I'm going to try to write the answer in a way I can add it as a comment to Hidden.tsx:

> Most components are automatically wrapped with a lot of boiler-plate form code in FormItem.tsx. Components that want more control over this can pass skipFieldDecorator and then use FormItem themselves. For hidden inputs, we want almost no wrapping at all so the component doesn't render anything on the page. To do this, we use skipFieldDecorator to opt out of boiler-plate then fieldConfigProp to get some advanced props ( like fieldConfig ) so that we can do some of what FormItem does for ourselves below:

Let me know if you have more questions that doesn't answer, or if something about the answer wasn't clear / could be clearer.

src/inputs/Hidden.tsx
ronaldzuniga: That's fine, I didn't notice they where in the inputs folder. That should be clear enough.

src/utilities/types.ts
karen-ka: That makes sense, thank you!

rkuykendall: Thanks!

PR #205, opened 2019-05-29
Add spacing under create fields on ObjectSearchCreate

rkuykendall:
Technical title: Wrap ObjectSearchCreate NestedFieldSet in Antd.Form.Item for spacing

This spacing issue was discovered while working on https://thatsmighty.atlassian.net/browse/MTY-886

## Screenshot before
image

## Screenshot after
image

### DIFF
--- a/src/building-blocks/NestedFieldSet.tsx
+++ b/src/building-blocks/NestedFieldSet.tsx
@@ -5,16 +5,11 @@ import autoBindMethods from 'class-autobind-decorator';
 
 import { splitName } from '@mighty-justice/utils';
 
-import {
-  fillInFieldSet,
-  FormFieldSet,
-  IFieldConfigPartial,
-  IFieldSetPartial,
-  mapFieldSetFields,
-} from '../';
-
-import { FormManager } from '../utilities';
+import { fillInFieldSet, FormManager, mapFieldSetFields } from '../utilities';
+import { IFieldConfigPartial, IFieldSetPartial } from '../interfaces';
 import { IModel } from '../props';
+
+import FormFieldSet from './FormFieldSet';
 
 export interface INestedFieldSetProps {
   fieldSet: IFieldSetPartial;
--- a/src/inputs/ObjectSearchCreate.tsx
+++ b/src/inputs/ObjectSearchCreate.tsx
@@ -96,23 +96,25 @@ class ObjectSearchCreate extends Component<IObjectSearchCreateProps> {
     const { fieldConfig, formManager } = this.injected;
 
     return (
-      <>
-        <NestedFieldSet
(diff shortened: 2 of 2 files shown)


src/building-blocks/NestedFieldSet.tsx
```diff
@@ -5,17 +5,12 @@ import autoBindMethods from 'class-autobind-decorator';
...
-import {
-  fillInFieldSet,
-  FormFieldSet,
-  IFieldConfigPartial,
-  IFieldSetPartial,
-  mapFieldSetFields,
-} from '../';
-
-import { FormManager } from '../utilities';
+import { fillInFieldSet, FormManager, mapFieldSetFields } from '../utilities';
```
artalpade: obviously looks better like this but why was it originally like line 8-14? What was the purpose?

src/inputs/ObjectSearchCreate.tsx
```diff
@@ -96,23 +96,25 @@ class ObjectSearchCreate extends Component<IObjectSearchCreateProps> {
...
-        />
-        <Antd.Button
-          className={`${CX_PREFIX_SEARCH_CREATE}-btn-back`}
-          onClick={this.onSearch}
-          size='small'
-        >
-          <Antd.Icon type='left' /> Back to search
-        </Antd.Button>
-      </>
+      <Antd.Col>
```
artalpade: what is the antd.col and why is it wrapped in it?

src/building-blocks/NestedFieldSet.tsx
rkuykendall: 8-14 just imports everything from `index.ts`, like so:

image

This is fine? I don't know. I don't feel strongly one way or the other. But Storybook has problems with imports like this, because it's not as smart as Babel/Webpack, so I switched the file to these separate imports.

src/inputs/ObjectSearchCreate.tsx
rkuykendall: I was mostly copying the wrapping provided by `FormItem.tsx` which wraps all our inputs:
image

`Col` is a Column layout component in Ant Design. I don't think it matters hugely here but I like the consistency with `FormItem`.

karen-ka: Did you add it just for consistency or also because it fixed the spacing issue?

rkuykendall: The spacing issue was fixed by `Form.Item`. The Col was added because it matched `FormItem.tsx` and I think having one be `Col > Form.Item` and the other being just `Form.Item` introduces an inconsistency I don't like.

karen-ka: Got it!

PR #206, opened 2019-05-29
Disable submit button on form validation errors

karen-ka:
Even though the form modal won't submit, the save button should be disabled if there is a custom validation error. 

JIRA: https://thatsmighty.atlassian.net/secure/RapidBoard.jspa?rapidView=31&projectKey=MTY&modal=detail&selectedIssue=MTY-470

Current: 

formbutton

### DIFF
--- a/src/components/Form.tsx
+++ b/src/components/Form.tsx
@@ -71,6 +71,10 @@ export class UnwrappedForm extends Component<IFormWrappedProps> {
     return fillInFieldSets(this.props.fieldSets);
   }
 
+  private hasErrors = (fieldsError: any) => {
+    return Object.keys(fieldsError).some((field) => fieldsError[field])
+  }
+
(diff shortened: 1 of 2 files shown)


src/components/Form.tsx
```diff
@@ -71,6 +71,9 @@ export class UnwrappedForm extends Component<IFormWrappedProps> {
     return fillInFieldSets(this.props.fieldSets);
   }
 
+  private hasErrors = (fieldsError: any) =>
```
ronaldzuniga: As much as I like this syntax, I think the standard of the project is to use `{}` and a return inside. @rkuykendall has the last word on this.

rkuykendall: I don't dislike this syntax tt all, but yeah, I think that we should keep things consistent with the rest of the project and expand this.

PR #208, opened 2019-05-30
Disable submit button when FormModal has errors on the page

karen-ka:
JIRA: https://thatsmighty.atlassian.net/secure/RapidBoard.jspa?rapidView=31&projectKey=MTY&modal=detail&selectedIssue=MTY-1160

modalform

### DIFF
--- a/test/features/submitButtonDisabled.test.tsx
+++ b/test/features/submitButtonDisabled.test.tsx
@@ -1,0 +1,37 @@
+import faker from 'faker';
+
+import { Tester } from '@mighty-justice/tester';
+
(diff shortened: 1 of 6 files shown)


test/features/submitButtonDisabled.test.tsx
```diff
@@ -0,0 +1,37 @@
+import faker from 'faker';
```
rpalermodrums: no need for this file to be `.tsx`. We only need that extension if there is jsx on the page (</>). Please rename to `submitButtonDisabled.test.ts`

Issue #189, opened 2019-05-13
403 forbidden should be caught by form manager


rkuykendall:
Example:
```
{"detail":"Only account admins may create or delete users."}
```

PR #212, opened 2019-06-05
Add descriptive error messages for 403 forbidden

karen-ka:
**JIRA**: https://thatsmighty.atlassian.net/secure/RapidBoard.jspa?rapidView=31&projectKey=MTY&modal=detail&selectedIssue=MTY-1149&assignee=karen

Github: https://github.com/mighty-justice/fields-ant/issues/189
Now: httpStatus.BAD_REQUEST is processed by FormManager and the useful messages presented to the user. 

These should also be presented to users, just like 400 errors.

The 'fix' . is probably just modifying the if statement to include both 400 and 403, but a test should be added for the use-case in the Github issue.

### DIFF
--- a/test/utilities/FormManager.test.tsx
+++ b/test/utilities/FormManager.test.tsx
@@ -1,15 +1,10 @@
 import faker from 'faker';
-import httpStatus from 'http-status-codes';
 import * as Antd from 'antd';
 
+import { ERROR_WITH_DESCRIPTION } from '../../src/utilities/FormManager';
 import { Tester } from '@mighty-justice/tester';
 
 import { Form, FormCard, IFieldSetPartial } from '../../src';
-
-const ERROR_TYPES = [
-  httpStatus.BAD_REQUEST,
-  httpStatus.FORBIDDEN,
-];
 
 async function getFormManager (fieldSets: IFieldSetPartial[], model = {}) {
   const props = {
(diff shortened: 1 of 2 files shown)


test/utilities/FormManager.test.tsx
```diff
@@ -80,4 +86,37 @@ describe('FormManager', () => {
        message: 'Error submitting form',
     });
   });
+
+  ERROR_TYPES.forEach(errorType => {
```
rkuykendall: 🙌

test/utilities/FormManager.test.tsx
```diff
@@ -1,9 +1,15 @@
 import faker from 'faker';
+import httpStatus from 'http-status-codes';
 import * as Antd from 'antd';
 
 import { Tester } from '@mighty-justice/tester';
 
-import { FormCard, IFieldSetPartial } from '../../src';
+import { Form, FormCard, IFieldSetPartial } from '../../src';
+
+const ERROR_TYPES = [
```
rkuykendall: - I feel like `ERROR_TYPES` is a bit too vague. These are the errors we can get data out of. What's a good name for them? `ERRORS_WITH_CONTENT`? `FEEDBACK_STATUS_CODES`? `ERROR_WITH_DESCRIPTION`? or anything you can come up with.
- If you added this const to formManager, you could write `describeErrors = ERROR_TYPES.includes(status);` (or whatever the better name is) which is nice.
- You could also export it and import it here and then we'd only have it written once.

PR #214, opened 2019-06-06
Bugfix: Submit is disabled when editing forms with nested objects

karen-ka:
JIRA: https://thatsmighty.atlassian.net/browse/MTY-1223

BEFORE:
bug-before

AFTER:

bug-after

### DIFF
--- a/src/utilities/FormManager.ts
+++ b/src/utilities/FormManager.ts
@@ -199,7 +199,7 @@ class FormManager {
 
   private hasErrors () {
     const fieldsError = flattenObject<{ [key: string]: any }, { [key: string]: any }>(this.form.getFieldsError());
-    return Object.keys((fieldsError)).some((field) => fieldsError[field]);
+    return Object.keys(fieldsError).some(field => fieldsError[field]);
   }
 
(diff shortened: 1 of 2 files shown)


src/utilities/FormManager.ts
```diff
@@ -197,8 +197,9 @@ class FormManager {
     });
   }
 
-  private hasErrors (fieldsError: any) {
-    return Object.keys(fieldsError).some((field) => fieldsError[field]);
+  private hasErrors () {
+    const fieldsError = flattenObject<{ [key: string]: any }, { [key: string]: any }>(this.form.getFieldsError());
```
rpalermodrums: kind of weird to read, but i suppose this is ok since it's taken from code earlier in the file

src/utilities/FormManager.ts
```diff
@@ -197,8 +197,9 @@ class FormManager {
...
+    return Object.keys((fieldsError)).some((field) => fieldsError[field]);
```
rpalermodrums: extra parentheses for both `fieldsError` and `field`

PR #216, opened 2019-06-08
Breaking change: ObjectSearchCreate takes full endpoint URL and handles errors

rkuykendall:
This change is actually for a side-project of mine that returns a 400 on empty searches. Right now that crashes ObjectSearch.

This will require very minor changes to places that use objectSearchCreate to re-add the / at the beginning and end.

### DIFF
--- a/src/inputs/ObjectSearch.tsx
+++ b/src/inputs/ObjectSearch.tsx
@@ -133,6 +133,9 @@ class ObjectSearch extends Component<IObjectSearchProps> {
     try {
       const response = await getEndpoint(`${endpoint}${toKey(params)}`);
       this.options = get(response, 'results', []);
+    } catch (error) {
+      // tslint:disable-next-line no-console
+      console.error(error);
     } finally {
       this.isLoading.setFalse();
     }
--- a/src/inputs/ObjectSearch.tsx
+++ b/src/inputs/ObjectSearch.tsx
@@ -130,9 +130,12 @@ class ObjectSearch extends Component<IObjectSearchProps> {
     this.search = value;
     this.isLoading.setTrue();
     this.updateValueCaches();
-    const response = await getEndpoint(`/${endpoint}/${toKey(params)}`);
-    this.options = response.results;
-    this.isLoading.setFalse();
+    try {
+      const response = await getEndpoint(`${endpoint}${toKey(params)}`);
+      this.options = get(response, 'results', []);
+    } finally {
+      this.isLoading.setFalse();
+    }
   }
 
   private renderOptionAdd () {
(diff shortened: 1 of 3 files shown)


src/inputs/ObjectSearch.tsx
```diff
@@ -130,9 +130,12 @@ class ObjectSearch extends Component<IObjectSearchProps> {
...
     this.isLoading.setTrue();
     this.updateValueCaches();
-    const response = await getEndpoint(`/${endpoint}/${toKey(params)}`);
-    this.options = response.results;
-    this.isLoading.setFalse();
+    try {
+      const response = await getEndpoint(`${endpoint}${toKey(params)}`);
+      this.options = get(response, 'results', []);
+    } finally {
+      this.isLoading.setFalse();
```
karen-ka: No errors are raised here if search fails?

test/types/objectSearch.test.ts
```diff
@@ -98,10 +98,10 @@ describe('objectSearch', () => {
   it('Properly caches and displays options', async () => {
     const { field, searchTerm, result, props, endpoint } = getComponentDefaults({})
       , tester = (await new Tester(ObjectSearch, { props }).mount() as any)
-      , newEndpoint = `${endpoint}-2`
+      , newEndpoint = `${endpoint}2/`
       ;
 
-    tester.endpoints[`/${newEndpoint}/`] = { results: [result] };
+    tester.endpoints[newEndpoint] = { results: [result] };
```
karen-ka: Was this change for style or was there a further technical purpose?

rkuykendall: Technical. Previously you would pass `lawfirms` and it would call getEndpoint with `/lawfirms/?search=` but if you want to do something like pass a full URL or some other more complicated endpoint, these slashes would be added no matter what. It doesn't help us much on the mighty end, but it makes this a much more limited component.

src/inputs/ObjectSearch.tsx
rkuykendall: Ah good point, no! Hmmm... how _should_ we handle that? Obviously, something like 5xx should maybe display a notification but... unsure about things like 4xx.

karen-ka: Maybe just display it if its 5xx?

rkuykendall: I started to implement toast notifications for errors, but upon thinking about it: In FormManager we make sure never to send toast notifications the library user can't stop or override. I think this would probably be a bad place to break that rule. But you're right, silently catching and squashing it is bad too. So instead I added a console.error, what do you think about that?

karen-ka: I think that's probably the best solution for now!

PR #217, opened 2019-06-17
Ability to easily add something to the extra of a form item an input

karen-ka:
**JIRA**: https://thatsmighty.atlassian.net/browse/MTY-1278

Use case: https://github.com/mighty-justice/registry-client/pull/654/files


rkuykendall:
A few questions about this:

1. Can you add this to the documentation?
2. How would you write this to do what we're doing in the reference, display the website of the currently-selected lawfirm?
3. If you don't pass anything, does it pass a <span></span> to extra, and is that a problem?
4. What if I want to show something generic, like 'Hey, have a nice day!' how would I do that?
5. Looks like this supports tooltip? Does extra handle that well?

API design is hard, but I'm glad we're figuring this out!

src/building-blocks/FormItem.tsx
```diff
@@ -69,18 +69,27 @@ class FormItem extends Component<IFormFieldProps> {
     };
   }
 
+  private get formItemExtra () {
+    const { fieldConfig } = this.props
+      , { formItemProps } = fieldConfig;
+    return (
+      <span>
+        { formItemProps && formItemProps.extraFieldConfig && renderLabel(formItemProps.extraFieldConfig) }
```
SHA-4: I would prefer this logic to be split out (e.g. return null if the first two aren't truthy)

src/building-blocks/FormItem.tsx
```diff
@@ -69,18 +69,27 @@ class FormItem extends Component<IFormFieldProps> {
     };
   }
 
+  private get formItemExtra () {
+    const { fieldConfig } = this.props
+      , { formItemProps } = fieldConfig;
+    return (
```
SHA-4: Space before return

src/building-blocks/FormItem.tsx
```diff
@@ -69,18 +69,27 @@ class FormItem extends Component<IFormFieldProps> {
...
+        { formItemProps && formItemProps.extraFieldConfig && renderLabel(formItemProps.extraFieldConfig) }
+      </span>
+    );
+  }
+
   public render () {
     const { formManager, fieldConfig } = this.props
       , { className, colProps, formItemProps, field } = fieldConfig
-      , { getFieldDecorator } = formManager.form
-      ;
```
SHA-4: No need for this syntax change

test/features/formItemExtra.test.ts
```diff
@@ -0,0 +1,31 @@
...
+  'FormModal',
+];
+
+// Tests that users can pass in optional extraFieldConfig props to render an extra under form item.
+describe('formItemExtra', () => {
+  SUPPORTING_COMPONENTS.forEach(componentName => {
+    it(`Form item extras correctly renders in ${componentName}`, async () => {
+      const { ComponentClass, propsFactory } = COMPONENT_GENERATORS[componentName]
+        , name = faker.lorem.sentence()
+        , name2 = faker.lorem.sentence()
```
SHA-4: Why include `name2` here?

SHA-4: Why include `name2` here?

test/features/formItemExtra.test.ts
```diff
@@ -0,0 +1,31 @@
...
+        , props = propsFactory.build({
+            fieldSets: [[{ field: 'name'
+            , formItemProps: { extraFieldConfig: {label: 'ExtraField', showLabel: true, tooltip: null} } }]],
```
SHA-4: Would you have to include `showLabel: true` and `tooltip: null` for this to work as intended? It seems like that should be the default

test/features/formItemExtra.test.ts
```diff
@@ -0,0 +1,31 @@
...
+            model: { name, name2 },
+        });
+
+      const tester = await new Tester(ComponentClass, { props }).mount();
+      expect(tester.find('.ant-form-extra').length).toBe(1);
```
SHA-4: I would want this to check if it includes the label

karen-ka:
@rkuykendall 

1. On it!
2. Good question, I think you could just do something like `extraFieldConfig: { field: 'case.law_firm.website', type: 'url' }`. I'll yalc it and try it out.
3. It does render an empty extra. I don't see it being a problem currently but for the long run that should definitely be changed. I'm not sure of a good way to overriding that attribute other than overriding the entire prop?
4. Not sure if I'm understanding this correctly, but to just render some plain text it would be something similar to the one in the test: `{ extraFieldConfig: {label: 'ExtraField', showLabel: true } }`.
5. It currently does, but now that I think about it that doesn't really make sense since extra should just be supplementing information.

src/building-blocks/FormItem.tsx
```diff
@@ -69,9 +69,31 @@ class FormItem extends Component<IFormFieldProps> {
     };
   }
 
+  private get getFieldConfig () {
```
SHA-4: When we write getters, we remove the get from the name. So it would be `private get fieldConfig`

src/building-blocks/FormItem.tsx
```diff
@@ -69,9 +69,31 @@ class FormItem extends Component<IFormFieldProps> {
...
+    const { fieldConfig, formModel } = this.props
+      , { field, formItemRenderExtra } = fieldConfig
+      , extraValue = get(formModel, field)
+      ;
+
+    if (!extraValue) { return fieldConfig; }
```
SHA-4: There are two instances where you return `fieldConfig`. You can combine them with some boolean logic

src/building-blocks/FormItem.tsx
```diff
@@ -69,9 +69,31 @@ class FormItem extends Component<IFormFieldProps> {
...
+
+    if (formItemRenderExtra) {
+
```
SHA-4: Unnecessary line here

src/building-blocks/FormItem.tsx
```diff
@@ -80,7 +102,7 @@ class FormItem extends Component<IFormFieldProps> {
         <Antd.Form.Item
           className={className}
           {...formItemProps}
```
rkuykendall: If we changed this to:
```ts
        <Antd.Form.Item
          className={className}
          {...this.formItemProps}
          {...formItemProps}
```
We could have a function that returns `{}` or `{ extra }` and then we wouldn't have to change fieldConafig at all. I think that would be a lot simpler than adding to formItemProps.

The reason I was doing it this way in the original was it was the only way for me to set extra from outside fields-ant, but you could just set it here yourself.

src/interfaces.ts
```diff
@@ -20,7 +20,8 @@ interface IFieldConfigBase {
   editProps: { [key: string]: any };
   field: string;
   fieldConfigProp: boolean;
-  formItemProps?: { [key: string]: any };
+  formItemProps?: Partial<FormItemProps>;
+  formItemRenderExtra?: (value: IValue) => IValue;
```
rkuykendall: This could probably return any React.ReactNode

SHA-4: Returning `IValue` is a little vague here. Wouldn't we expect it to be a string or react node?

test/features/formItemExtra.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+];
+
+function testFormat (value: IValue) {
+    if (value) {
+      return <span id="test">Hello</span>
+    }
+    return null;
+}
+
+// Tests that users can pass in optional extraFieldConfig props to render an extra under form item.
```
SHA-4: What are you referring to with `extraFieldConfig props`?

test/features/formItemExtra.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { IValue } from '../../src/props';
+
+const SUPPORTING_COMPONENTS = [
+  'Form',
+  'FormCard',
+  'FormDrawer',
+  'FormModal',
+];
+
+function testFormat (value: IValue) {
```
SHA-4: This formatter can be named `extraRenderer` or something along those lines for clarity's sake (it could also be defined inline)

test/features/formItemExtra.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+      const tester = await new Tester(ComponentClass, { props }).mount();
+      expect(tester.find('.ant-form-extra').length).toBe(1);
+      expect(tester.find('.ant-form-extra #test').text()).toEqual("Hello");
+    });
+
+    it(`No form items extras render if no props passed in`, async () => {
+        const { ComponentClass, propsFactory } = COMPONENT_GENERATORS[componentName]
+          , name = faker.lorem.sentence()
+          , props = propsFactory.build({
+              fieldSets: [[{ field: 'name', model: { name} } ]],
```
SHA-4: Do you have to do this much lift (e.g. passing faking name, passing model)? Also, spacing in `{ name}`.

test/features/formItemExtra.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+}
+
+// Tests that users can pass in optional extraFieldConfig props to render an extra under form item.
+describe('formItemExtra', () => {
+  SUPPORTING_COMPONENTS.forEach(componentName => {
+    it(`Form item extras correctly renders in ${componentName}`, async () => {
+      const { ComponentClass, propsFactory } = COMPONENT_GENERATORS[componentName]
+        , name = faker.lorem.sentence()
+        , props = propsFactory.build({
+            fieldSets: [[{ field: 'name'
```
SHA-4: Formatting for the brackets of `fieldSets` is a little off. I would put attributes on their own lines without brackets

test/features/formItemExtra.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+            , formItemRenderExtra: testFormat }]],
+            model: { name },
+        });
+
+      const tester = await new Tester(ComponentClass, { props }).mount();
```
SHA-4: Right here, you are awaiting but not mounting tester with `async: true`

test/features/formItemExtra.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+    });
+
+    it(`No form items extras render if no props passed in`, async () => {
+        const { ComponentClass, propsFactory } = COMPONENT_GENERATORS[componentName]
+          , name = faker.lorem.sentence()
+          , props = propsFactory.build({
+              fieldSets: [[{ field: 'name', model: { name} } ]],
+          });
+
+        const tester = await new Tester(ComponentClass, { props }).mount();
```
SHA-4: While it is clear to set the tester const like this, it's nice to append it to the const declarations beforehand as well

test/features/formItemExtra.test.tsx
karen-ka: I just checked, elsewhere in the codebase it's all on one line. Should I follow that or separate onto new lines?

SHA-4: Either way, just not halfway!

README.md
```diff
@@ -78,6 +78,7 @@ an appropriate render function, edit component, validator, and sets falsey value
 
 #### Core attributes
 - **className**: string, applied on the `Antd.Form.Item`.
+- **formItemRenderExtra**: Optional function for rendering an extra under a form item. See interfaces.ts for call signature.
```
rkuykendall: 🙌

test/features/formItemExtra.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { IValue } from '../../src/props';
+
+const SUPPORTING_COMPONENTS = [
+  'Form',
+  'FormCard',
+  'FormDrawer',
+  'FormModal',
+];
+
+function extraRenderer (value: IValue) {
```
SHA-4: Can shorten this rendered significantly with a ternary: `return value ? <span>...</span> : null;`

test/features/formItemExtra.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+        , tester = await new Tester(ComponentClass, { props }).mount({ async: true });
+
+      expect(tester.find('.ant-form-extra').length).toBe(1);
+      expect(tester.find('.ant-form-extra #test').text()).toEqual("Hello");
+    });
+
+    it(`No form items extras render if no props passed in`, async () => {
+        const { ComponentClass, propsFactory } = COMPONENT_GENERATORS[componentName]
+          , props = propsFactory.build()
+          , tester = await new Tester(ComponentClass, { props }).mount();
```
SHA-4: You also await without async here :)

test/features/formItemExtra.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+const SUPPORTING_COMPONENTS = [
+  'Form',
+  'FormCard',
+  'FormDrawer',
+  'FormModal',
+];
+
+function extraRenderer (value: IValue) {
+    if (value) {
+      return <span id="test">Hello</span>
```
SHA-4: Looks like we use single quotes rather than double in `fields-ant` (also see "Hello" below)

test/features/formItemExtra.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+// Tests that users can pass in optional formItemRenderExtra to render an extra under form item.
+describe('formItemExtra', () => {
+  SUPPORTING_COMPONENTS.forEach(componentName => {
+    it(`Form item extras correctly renders in ${componentName}`, async () => {
+      const { ComponentClass, propsFactory } = COMPONENT_GENERATORS[componentName]
+        , name = faker.lorem.sentence()
+        , props = propsFactory.build({
+            fieldSets: [[{ field: 'name', formItemRenderExtra: extraRenderer }]],
+            model: { name }
```
SHA-4: trailing comma

PR #222, opened 2019-06-26
Update README with yarn commands

rkuykendall:


README.md
```diff
@@ -226,3 +226,15 @@ Gotchas:
 
 - Make sure to `npm install` or `yarn install` first
 - The deploy command must currently be run with npm not yarn due to login bug
+
```
artalpade: Could be more explicit but I guess it's okay...

README.md
```diff
@@ -226,3 +226,15 @@ Gotchas:
...
+# All other commands
+
+- **yarn build** - Build the project
+- **yarn dev** - Build the project and watch for changes
+- **yarn docs:build** - Build storybook documentation
+- **yarn docs:deploy** - Deploy storybook documentation to Github pages
```
karen-ka: not sure if it's relevant but a note about turning off deploy when yalc-ing so you don't get spammed ?

rkuykendall: I feel like that would be relevant in a section we should maybe add on Yalcing 🤔

PR #223, opened 2019-07-05
Change add new text to more closely match usage

rkuykendall:
**Jira:** https://thatsmighty.atlassian.net/browse/MTY-1310

image

### DIFF
--- a/src/inputs/ObjectSearch.tsx
+++ b/src/inputs/ObjectSearch.tsx
@@ -149,6 +149,7 @@ class ObjectSearch extends Component<IObjectSearchProps> {
 
   private renderOptionAdd () {
     const { addNewContent } = this.props
+      , { label } = this.fieldConfig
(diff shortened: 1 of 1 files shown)


src/inputs/ObjectSearch.tsx
```diff
@@ -161,7 +162,7 @@ class ObjectSearch extends Component<IObjectSearchProps> {
 
     return (
       <Antd.Select.Option className={className} key={ITEM_KEYS.ADD}>
-        <div>{addNewContent || <><Antd.Icon type='plus' /> Add new</>}</div>
+        <div>{addNewContent || <>Can't find it? <a>Add new {label}</a></>}</div>
```
artalpade: So before it'd just be a place to add a new object search create field without a label?

rkuykendall: Before it did the same thing: It's the button to say none of these search results are what I want, so I want to create a new one.

The text is just slightly different. Kevin in Registry kept overriding this text instead of just fixing it in the library. This will allow me to delete those overrides.

PR #224, opened 2019-07-08
Add address type

SHA-4:


src/inputs/Address.tsx
```diff
@@ -0,0 +1,83 @@
...
+  public render () {
+    const { fieldConfig, formManager } = this.injected;
+
+    return (
+      <Antd.Col>
+        <Antd.Form.Item className={fieldConfig.className}>
+          <NestedFieldSet
+            fieldSet={this.fieldSet}
+            formManager={formManager}
+            formModel={formManager.formModel}
```
rkuykendall: You should use the formModel you are passed from the props. The whole point of passing it around is that accessing formManager.formModel is extremely costly and we want to try to do it high-up once and pass it down.

src/utilities/common.tsx
```diff
@@ -130,6 +132,28 @@ export function fieldSetsToColumns (fieldSets: IFieldSetPartial[], tableModel: I
...
+  ;
+
+  // When using the add new feature of objectSearchCreate, we should
+  // make sure to nullify the appropriate fields in the new model
+  if (isTypeObjectSearchCreate(fieldConfig) && isAddingNew) {
+    return modelFromFieldConfigs(getFieldSetFields(fieldConfig.createFields).map(fillInFieldConfig), formValue);
+  }
+
+  // When saving an address, nullify if no attributes
+  if (isTypeAddress(fieldConfig) && !some(formValue)) {
```
rkuykendall: Just a note that empty object is truthy. You _may_ want to consider _.isEmpty if that applies to you.

src/utilities/common.tsx
```diff
@@ -142,18 +166,8 @@ export function modelFromFieldConfigs (fieldConfigs: IFieldConfig[], data: IMode
...
-
-          // When using the add new feature of objectSearchCreate, we should
-          // make sure to nullify the appropriate fields in the new model
-          , isAddingNew = isObject(formValue) && !has(formValue, ID_ATTR)
-          , value = (isTypeObjectSearchCreate(fieldConfig) && isAddingNew)
-            ? modelFromFieldConfigs(getFieldSetFields(fieldConfig.createFields).map(fillInFieldConfig), formValue)
-            : nullifiedValue
-            ;
+        const { field } = fieldConfig
+          , value = nullifyValue(fieldConfig, data);
```
rkuykendall: 🙌

src/interfaces.ts
```diff
@@ -66,6 +66,10 @@ export interface IFieldConfigOptionSelect extends IFieldConfigBase {
   sorted?: boolean;
 }
 
+export interface IFieldConfigAddress extends IFieldConfigBase {
+  stateProps: IFieldConfigOptionSelect;
```
rkuykendall: Since optionSelect is a full fieldConfig, and this does not require `field` ( plus field makes no sense here ), perhaps it should be Partial<IFieldConfigOptionSelect>? Also since this won't be filled in in the `fillInFieldConfig` it's actually `stateProps?:` right?

PR #225, opened 2019-07-25
Failing test case for date type error

artalpade:
https://thatsmighty.atlassian.net/browse/MTY-1313

Wrote a failing test case for the fieldsant error where you cannot submit a form after typing a date and then deleting it even if it's optional

### DIFF
--- a/test/types/date.test.ts
+++ b/test/types/date.test.ts
@@ -86,4 +86,26 @@ describe('date', () => {
     expect(await checkInferYear('50')).toBe('1950');
     expect(await checkInferYear('98')).toBe('1998');
   });
+
+
+  it('Changes input before submission', async () => {
+  const { fieldConfigFactory } = TYPE_GENERATORS.date
+    , fieldConfig = fieldConfigFactory.build()
+    , fieldSets = [[fieldConfig]]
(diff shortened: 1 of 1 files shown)


test/types/date.test.ts
```diff
@@ -86,4 +86,26 @@ describe('date', () => {
...
+  //enter normal date
+  tester.changeInput(`input[id="${fieldConfig.field}.month"]`, '01');
+  tester.changeInput(`input[id="${fieldConfig.field}.day"]`, '01');
+  tester.changeInput(`input[id="${fieldConfig.field}.year"]`, '2019');
+  //clear all fields
+  tester.changeInput(`input[id="${fieldConfig.field}.month"]`, '');
+  tester.changeInput(`input[id="${fieldConfig.field}.day"]`, '');
+  tester.changeInput(`input[id="${fieldConfig.field}.year"]`, '');
+  await tester.submit();
+  expect(await !!onSave.mock.calls.length).toBe(true);
```
rkuykendall: What is this await for???

artalpade: All the other expect() statements had await in them so I assumed it was needed

PR #230, opened 2019-08-08
Infer state in address type

karen-ka:
Add ability to infer state from zipcode if `smart` is set to true

### DIFF
--- a/package.json
+++ b/package.json
@@ -146,7 +146,8 @@
     "storybook-readme": "^5.0.1",
     "style-loader": "^0.23.1",
     "tslint": "5.14.0",
-    "typescript": "3.3.4000"
+    "typescript": "3.3.4000",
+    "zipcodes": "^8.0.0"
   },
   "dependencies": {
     "@mighty-justice/smart-bool": "^1.0.0",
--- a/package.json
+++ b/package.json
@@ -152,6 +152,7 @@
   "dependencies": {
     "@mighty-justice/smart-bool": "^1.0.0",
     "@mighty-justice/utils": "1.3.4",
+    "@types/zipcodes": "^6.1.0",
     "class-autobind-decorator": "^3.0.1",
     "classnames": "^2.2.6",
     "date-fns": "^1.30.1",
--- a/package.json
+++ b/package.json
@@ -114,6 +114,7 @@
     "@types/storybook__addon-actions": "^3.4.2",
     "@types/storybook__addon-info": "^4.1.1",
     "@types/storybook__react": "^4.0.1",
+    "@types/zipcodes": "^6.1.0",
     "antd": "^3.0.0",
     "auto-changelog": "^1.12.0",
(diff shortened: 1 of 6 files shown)


src/global.d.ts
```diff
@@ -2,3 +2,5 @@ declare module '*.md' {
     const content: string;
     export default content;
 }
+
+declare module 'zipcodes';
```
rkuykendall: `yarn add @types/zipcodes` and remove this to get type information.

test/types/address.test.ts
```diff
@@ -33,4 +33,16 @@ describe('address', () => {
     tester.submit();
     expect(onSave).toHaveBeenCalledWith({[fieldConfig.field]: emptyValue});
   });
+
+  it('Automatically fills in state if smart mode is on', async () => {
+    const { fieldConfigFactory, valueFunction } = TYPE_GENERATORS.address
+      , value = valueFunction()
+      , addressValues = mapValues(value, () => '02912')
```
rkuykendall: Shouldn't only zip be `02912 ` and not every field?

src/inputs/Address.tsx
```diff
@@ -60,16 +62,29 @@ class Address extends Component<IAddressProps> {
     return fieldSet.map((addressConfig: IFieldConfigPartial) => ({ ...addressConfig, colProps }));
   }
 
-  public render () {
-    const { fieldConfig, formManager, formModel } = this.injected;
+  private get model () {
+    const { fieldConfig: { field, smart }, formModel } = this.injected
+      , zip = get(formModel[field], 'zip_code');
+
+    if (zip && smart) {
```
rkuykendall: Should this also check that there ISN'T a state?

package.json
```diff
@@ -146,11 +146,13 @@
...
     "style-loader": "^0.23.1",
     "tslint": "5.14.0",
-    "typescript": "3.3.4000"
+    "typescript": "3.3.4000",
+    "zipcodes": "^8.0.0"
   },
   "dependencies": {
     "@mighty-justice/smart-bool": "^1.0.0",
     "@mighty-justice/utils": "1.3.4",
+    "@types/zipcodes": "^6.1.0",
```
rkuykendall: This should be moved to devDependencies

test/types/address.test.ts
```diff
@@ -33,4 +33,18 @@ describe('address', () => {
...
+      , value = valueFunction()
+      , addressValues = mapValues(value, () => '')
+      , fieldConfig = { ...fieldConfigFactory.build(), smart: true }
+    ;
+
+    addressValues['zip_code'] = '02912';
+
+    const props = { fieldSets: [[fieldConfig]], model: { [fieldConfig.field]: addressValues } }
+      , tester = await new Tester(FormCard, {props}).mount();
+    expect(tester.html()).toContain('Rhode Island');
```
rkuykendall: Can you submit and check the submit data as well?

PR #241, opened 2019-09-18
Add support for disabled param to ObjectSearchCreate options

wpontius:
See https://ant.design/components/select/#Option-props

### DIFF
--- a/src/inputs/ObjectSearch.tsx
+++ b/src/inputs/ObjectSearch.tsx
@@ -24,6 +24,7 @@ export interface IObjectSearchProps {
   addNewContent?: React.ReactNode;
   debounceWait: number;
   fieldConfig: IFieldConfigObjectSearchCreate;
+  isOptionDisabled?: (option: IEndpointOption) => boolean;
   loadingIcon?: React.ReactNode;
   noSearchContent?: React.ReactNode;
   onAddNew?: (search: string) => void;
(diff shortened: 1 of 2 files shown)


src/inputs/ObjectSearch.tsx
```diff
@@ -190,12 +190,13 @@ class ObjectSearch extends Component<IObjectSearchProps> {
...
 
   private renderOption (option: IEndpointOption) {
-    const { renderOption, renderSelected } = this.fieldConfig
+    const { isDisabled, renderOption, renderSelected } = this.fieldConfig
        , className = `${CX_PREFIX_SEARCH_CREATE}-item`;
 
     return (
       <Antd.Select.Option
         className={className}
+        disabled={isDisabled ? isDisabled(option) : false}
```
rkuykendall: I would make it `isOptionDisabled` and instead of adding it to the fieldConfig, add it to the component props, then your fieldConfig can contain `{ editProps: { isOptionDisabled } }` and we don't need to change the fieldConfig definition.

src/inputs/ObjectSearch.tsx
```diff
@@ -190,12 +190,17 @@ class ObjectSearch extends Component<IObjectSearchProps> {
   }
 
   private renderOption (option: IEndpointOption) {
-    const { renderOption, renderSelected } = this.fieldConfig
-       , className = `${CX_PREFIX_SEARCH_CREATE}-item`;
+    const {
+        editProps : { isOptionDisabled },
```
rkuykendall: editProps get passed right into the component, so you can add `isOptionDisabled?:` to `IObjectSearchProps` and  `{ IObjectSearchProps } = this.props`

PR #243, opened 2019-09-20
Fix ObjectSearch multiselect

mumumumu:
Store previously selected options because options get cleared on each search.

### DIFF
--- a/src/inputs/ObjectSearch.tsx
+++ b/src/inputs/ObjectSearch.tsx
@@ -50,7 +50,6 @@ class ObjectSearch extends Component<IObjectSearchProps> {
 
   @observable private previousEndpoint = '';
   @observable private previousSearchFilters = '';
-  @observable private previousSelectedOptions: IEndpointOption[] = [];
 
   public static defaultProps: Partial<IObjectSearchProps> = {
     debounceWait: DEFAULT_DEBOUNCE_WAIT,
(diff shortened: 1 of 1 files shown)


src/inputs/ObjectSearch.tsx
```diff
@@ -233,7 +234,9 @@ class ObjectSearch extends Component<IObjectSearchProps> {
     // Select from search
     if (this.isMultiSelect) {
       const selectedOptionIds = selectedOption.map((_selectedOption: any) => _selectedOption.key)
-        , foundOptions = this.options.filter(option => selectedOptionIds.includes(option.id));
+        , optionsToSearch = Array.from(new Set([...this.previousSelectedOptions, ...this.options]))
+        , foundOptions = optionsToSearch.filter((option: any) => selectedOptionIds.includes(option.id));
+      this.previousSelectedOptions = foundOptions;
```
rkuykendall: Would this break on a non-empty form? Aren't `previousSelectedOptions ` in `value`?

mumumumu: Yeah, that's a good catch, will update.

src/inputs/ObjectSearch.tsx
```diff
@@ -233,7 +233,8 @@ class ObjectSearch extends Component<IObjectSearchProps> {
     // Select from search
     if (this.isMultiSelect) {
       const selectedOptionIds = selectedOption.map((_selectedOption: any) => _selectedOption.key)
-        , foundOptions = this.options.filter(option => selectedOptionIds.includes(option.id));
+        , optionsToSearch = Array.from(new Set([...this.injected.value, ...this.options]))
```
rkuykendall: Since we don't have an ORM, I don't _think_  `Array.from(new Set(` here will actually remove duplicates. It would keep both, as shown here:
```ts
Array.from(new Set([{ id: '1' }, { id: '1' }])) => [{ id: '1' }, { id: '1' }]
```
Perhaps instead we should do use [uniqBy](https://lodash.com/docs/4.17.15#uniqBy)?
```ts
_.uniqBy([{ id: '1' }, { id: '1' }], 'id').length => 1
```
like so:
```ts
        , optionsToSearch = uniqBy([...this.injected.value, ...this.options], 'id')
```

PR #244, opened 2019-09-25
Fix to pass isOptionDisabled in ObjectSearchCreate edit props

wpontius:
Follow up to #241 .

To be passed as an edit prop, the param `isOptionDisabled` must be added to the OSC interface and specifically picked from `editProps`.

### DIFF
--- a/src/inputs/ObjectSearchCreate.tsx
+++ b/src/inputs/ObjectSearchCreate.tsx
@@ -14,6 +14,7 @@ import {
   CX_PREFIX_SEARCH_CREATE,
   FormManager,
   IAntFormField,
+  IEndpointOption,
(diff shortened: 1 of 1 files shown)


src/inputs/ObjectSearchCreate.tsx
```diff
@@ -42,6 +42,7 @@ export interface IObjectSearchCreateProps {
   fieldDecorator: <T>(component: T) => T;
   formManager: FormManager;
   formModel: IModel;
+  isOptionDisabled?: (option: any) => boolean;
```
SHA-4: Replace the `any` with `IEndpointOption`? That's what the object search props indicates

PR #248, opened 2019-11-05
Add trim whitespace component and apply to url and email fields

rpalermodrums:
If you copy and paste an email/website in the email/website fields and there is a space, it won’t let you save the email/website. This PR trims the spaces on both fields since there should never be whitespace on either

Will share Ngrok link upon request

### DIFF
--- a/test/types/trimWhitespaceInput.test.ts
+++ b/test/types/trimWhitespaceInput.test.ts
@@ -1,0 +1,35 @@
+import * as faker from 'faker';
+
+import { Tester } from '@mighty-justice/tester';
+
+import { TYPE_GENERATORS, valueRenderPairs } from '../factories';
+
+import { FormCard } from '../../src';
+
+const { fieldConfigFactory } = TYPE_GENERATORS['email']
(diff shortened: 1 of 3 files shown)


test/types/trimWhitespaceInput.test.ts
```diff
@@ -0,0 +1,35 @@
...
+
+describe('TrimWhitespaceInput', () => {
+  it('trims whitespace', async () => {
+    const onSave = jest.fn()
+      , props = { fieldSets, model, onSave }
+      , fakeEmail = faker.internet.email()
+      , emailWithSpace = `       ${fakeEmail}      `
+      ;
+
+    // Renders formatted value
```
SHA-4: nit: unnecessary comment

test/types/trimWhitespaceInput.test.ts
```diff
@@ -0,0 +1,35 @@
...
+    const tester = await new Tester(FormCard, { props }).mount();
+    expect(onSave).not.toHaveBeenCalled();
```
SHA-4: nit: spacing between expects and actions

PR #249, opened 2019-12-11
Fix Checkbox not showing as reset after submit

rkuykendall:
Test failing if you remove Checkbox.tsx changes:
image

### DIFF
--- a/src/inputs/Checkbox.tsx
+++ b/src/inputs/Checkbox.tsx
@@ -2,20 +2,30 @@ import React, { Component } from 'react';
 import { observer } from 'mobx-react';
 import autoBindMethods from 'class-autobind-decorator';
 import { omit } from 'lodash';
-import { ICheckboxProps } from '../interfaces';
 
 import * as Antd from 'antd';
 
-@autoBindMethods()
+import { IAntFormField, ICheckboxProps, IInjected } from '../interfaces';
+
+@autoBindMethods
 @observer
 export default class Checkbox extends Component <ICheckboxProps> {
+  private get injected () {
+    return this.props as IInjected & ICheckboxProps & IAntFormField;
+  }
+
   public render () {
-    const { description, disabled, disabledText } = this.props;
+    const { description, disabled, disabledText, value, id } = this.injected
+      , PROPS_TO_OMIT = ['description', 'value']
+      , checkboxProps = {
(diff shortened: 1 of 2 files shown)


src/inputs/Checkbox.tsx
```diff
@@ -2,20 +2,30 @@ import React, { Component } from 'react';
 import { observer } from 'mobx-react';
 import autoBindMethods from 'class-autobind-decorator';
 import { omit } from 'lodash';
-import { ICheckboxProps } from '../interfaces';
 
 import * as Antd from 'antd';
 
-@autoBindMethods()
+import { IAntFormField, ICheckboxProps, IInjected } from '../interfaces';
+
```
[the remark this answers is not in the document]
rkuykendall: No, that's a good question, I was sorta on the fence. I left it for the same reason that I leave `autobindmethods` on a component even when I don't technically need it: We save a lot of developer time and sanity through consistency. See the other components that use `this.injected`:
image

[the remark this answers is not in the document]
ekatsuta: got it. thanks!

test/types/checkbox.test.ts
```diff
@@ -0,0 +1,43 @@
+import { Tester } from '@mighty-justice/tester';
+
+import { TYPE_GENERATORS } from '../factories';
+
```
wsaffran: Is the space between these imports standard?

rkuykendall: Yeah, we usually put blank spaces between:
```
global imports
ant-design imports
@mighty-justice imports
../ imports
./ imports
style.module.less import
```

test/types/checkbox.test.ts
```diff
@@ -0,0 +1,43 @@
...
+describe('checkbox', () => {
+  it('resets on submit', async () => {
+    const onSave = jest.fn()
+      , props: Partial<IFormCardProps> = { fieldSets, onSave, resetOnSuccess: true };
+
+    const tester = await new Tester(FormCard, { props }).mount();
+
+    async function submitAndCheckFor (value: boolean) {
+      expect(onSave).not.toHaveBeenCalled();
+      expect(tester.find('.ant-checkbox-wrapper-checked').length).toBe(value ? 1 : 0);
```
SHA-4: You can use `exists()` to remove the ternary

PR #250, opened 2019-12-11
Add searchOnEmpty prop to ObjectSearch and OSC

rkuykendall:
# Screenshots
**Before:**
osc-before

**After:**
osc-after

### DIFF
--- a/src/inputs/ObjectSearch.tsx
+++ b/src/inputs/ObjectSearch.tsx
@@ -137,7 +137,10 @@ class ObjectSearch extends Component<IObjectSearchProps> {
       ;
 
     this.search = value;
-    if (!searchOnEmpty && !this.hasSearch) { return; }
+    if (!searchOnEmpty && !this.hasSearch) {
+      this.options = [];
+      return;
+    }
 
     this.isLoading.setTrue();
     this.updateValueCaches();
@@ -286,7 +289,8 @@ class ObjectSearch extends Component<IObjectSearchProps> {
   public render () {
     const { id, onAddNew, searchOnEmpty } = this.injected
       , isLoading = this.isLoading.isTrue
-      , showNoResultsOption = (this.hasSearch || searchOnEmpty) && !isLoading && !this.hasOptions
+      , canSearch = this.hasSearch || searchOnEmpty
+      , showNoResultsOption = canSearch && !isLoading && !this.hasOptions
       , showAddOption = this.hasSearch && onAddNew
       , showNoSearch = !this.hasSearch
       , { label, showLabel } = this.fieldConfig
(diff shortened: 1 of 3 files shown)


src/inputs/ObjectSearch.tsx
```diff
@@ -136,6 +137,8 @@ class ObjectSearch extends Component<IObjectSearchProps> {
       ;
 
     this.search = value;
+    if (!searchOnEmpty && !this.hasSearch) { return; }
```
SHA-4: This makes it so if you type and delete, it continues to show the results from the typed search if `!searchOnEmpty`. It's minor but assumedly unintentional?

test/types/objectSearch.test.ts
```diff
@@ -127,4 +127,27 @@ describe('objectSearch', () => {
     expect(tester.text()).not.toContain(results[1].name);
     expect(tester.text()).toContain(results[2].name);
   });
+
+  [true, false].forEach((searchOnEmpty: boolean) => {
+    it(`Works for searchOnEmpty={${searchOnEmpty}} prop`, async () => {
+      const { field, searchTerm, results, props } = getComponentDefaults({})
+        , tester = (await new Tester(ObjectSearch, { props: { ...props, searchOnEmpty }}).mount() as any)
+        , callsOnFocus: number = searchOnEmpty ? 1 : 0
```
SHA-4: This is clear, but you could use `+searchOnEmtpy` instead of a ternary

src/inputs/ObjectSearch.tsx
```diff
@@ -278,19 +284,19 @@ class ObjectSearch extends Component<IObjectSearchProps> {
   }
 
   public render () {
-    const { id, onAddNew } = this.injected
+    const { id, onAddNew, searchOnEmpty } = this.injected
       , isLoading = this.isLoading.isTrue
-      , showNoResultsOption = this.hasSearch && !this.hasOptions
+      , showNoResultsOption = (this.hasSearch || searchOnEmpty) && !isLoading && !this.hasOptions
```
SHA-4: This boolean logic is a little complex. Can you split it into multiple variables?

test/types/objectSearch.test.ts
rkuykendall: You are a bad person.

src/inputs/ObjectSearch.tsx
rkuykendall: Better?
```ts
      , canSearch = this.hasSearch || searchOnEmpty
      , showNoResultsOption = canSearch && !isLoading && !this.hasOptions
```

src/inputs/ObjectSearch.tsx
rkuykendall: :( yeah I'll work on that

PR #252, opened 2019-12-13
Add test for nullification, fix datepicker nullification

rkuykendall:


test/types/types.test.ts
```diff
@@ -1,8 +1,9 @@
 import { Tester } from '@mighty-justice/tester';
 import { EMPTY_FIELD } from '@mighty-justice/utils';
 
-import { Card, fillInFieldConfig, FormCard, TYPES } from '../../src';
+import { Card, fillInFieldConfig, FormCard, IFieldConfig, TYPES } from '../../src';
 import { TYPE_GENERATORS, valueRenderPairs } from '../factories';
+import { IValue } from '../../src/props';
```
ekatsuta: is this similar to importing IInjected for possible future use case? IValue not used anywhere else

rkuykendall: NOPE! This is a mistake you found :)

test/types/types.test.ts
```diff
@@ -48,6 +48,22 @@ Object.keys(TYPE_GENERATORS).forEach(type => {
...
+      const emptyValue = (() => {
+          if (fieldConfig.type === 'checkbox') { return false; }
+          return fieldConfig.nullify ? null : '';
+        })()
+        , onSave = jest.fn()
+        , props = { fieldSets, onSave }
+        , expectModel = { [fieldConfig.field]: emptyValue }
+        ;
+
+      const tester = await new Tester(FormCard, { props }).mount();
```
SHA-4: This works because it mounts quickly, but shouldn't you pass `async: true`?

test/types/types.test.ts
```diff
@@ -48,6 +48,22 @@ Object.keys(TYPE_GENERATORS).forEach(type => {
       expect(onSave).toHaveBeenCalledWith(model);
     });
 
+    it('Has correct empty values', async () => {
+      const emptyValue = (() => {
```
SHA-4: I would use a named function for this that takes a `fieldConfig` prop. I can imagine this being useful in other tests

PR #257, opened 2020-01-15
Add new objectSelect input type

rkuykendall:
This is like OptionSelect, but it doesn't force the shape to be `{ name, value }` and in fact defaults to `getNameOrDefault` and `id`. This was created because we needed to have the user select from a list of objects ( `ExternalSystem` objects ) and it didn't fit cleanly into OptionSelect or ObjectSearch, the latter of which is very API focused, and the former of which is very regimented to the IOption shape.

image


src/inputs/ObjectSelect.tsx
```diff
@@ -0,0 +1,94 @@
...
+import {
+  IAntFormField,
+  IFieldConfigOptionSelect,
+  IInjected,
+  IInputProps,
+} from '../interfaces';
+
+import { getOptions } from '../utilities';
+import { IModel } from '../props';
+import { getNameOrDefault } from '@mighty-justice/utils';
```
wsaffran: Import order

src/props.ts
```diff
@@ -26,7 +26,6 @@ export { ISummaryCardProps } from './components/SummaryCard';
 // Form inputs
 export { IObjectSearchCreateProps } from './inputs/ObjectSearchCreate';
 export { IOptionSelectDisplayProps } from './inputs/OptionSelectDisplay';
-export { IOptionSelectProps } from './inputs/OptionSelect';
```
SHA-4: Why not export `IObjectSelectProps` here?

src/inputs/OptionSelect.tsx
```diff
@@ -1,76 +1,16 @@
...
-  private get showSearch (): boolean {
-    // the showSearch fieldConfig option will override this
-    if (isBoolean(this.fieldConfig.showSearch)) {
-      return this.fieldConfig.showSearch;
-    }
-
-    return this.options.length > SHOW_OPTION_SEARCH_IF_OVER;
-  }
-
+class OptionSelect extends Component<IInputProps> {
```
SHA-4: Since you set this to receive `IInputProps`, won't this result in a type error if you pass anything in `IObjectSelectProps` to this component?

src/inputs/ObjectSelect.tsx
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  public static defaultProps: Partial<IObjectSelectProps> = {
+    keyBy: 'id',
+    renderOption: getNameOrDefault,
+    renderSelected: getNameOrDefault,
+  };
+
+  private get fieldConfig () {
+    return this.props.fieldConfig as IFieldConfigOptionSelect;
```
SHA-4: This interface name looks outdated

rkuykendall: TBH it's complicated. I can go over it in person. Long story short it is outdated, and even the concept of an option is outdated, and I tried to fix it and this PR got 3x larger and everything broke, so I backed off.

SHA-4: If you can make an issue about it, I'm happy with this for now

PR #260, opened 2020-01-30
Bugfix: Form remains disabled after fixing backend validation

rkuykendall:


src/building-blocks/FormItem.tsx
```diff
@@ -46,20 +46,23 @@ class FormItem extends Component<IFormFieldProps> {
...
-        // Our own proprietary ( much more sane and powerful ) validation attribute
-        // is converted here to the rc-form style validator
-        if (validationRule.fieldsValidator) {
-          return {
-            validator: this.fieldsValidatorToValidator(validationRule.fieldsValidator, validationRule.message),
-            ...omit(validationRule, 'fieldsValidator'),
-          };
-        }
+    return [
+      { validator: noopValidator },
```
ekatsuta: more of a question...so does this ensure that every time the `rules()` is called, we clear the backend errors?  If so, we can't directly call the `callback()` here?

rkuykendall: It ensures that when a field is edited the validation is re-run and so it clears the manually set backend validation. I should add a note here.

rkuykendall: Added

PR #273, opened 2020-03-05
Fix disable attribute of fieldConfig

SHA-4:
Added support for the disabled attribute for all types

### DIFF
--- a/src/building-blocks/FormField.tsx
+++ b/src/building-blocks/FormField.tsx
@@ -29,6 +29,7 @@ class FormField extends Component<IFormFieldProps> {
       ;
 
     return {
+      disabled: fieldConfig.disabled,
       ...fieldConfig.editProps,
       ...fieldConfigProp,
     };
--- a/src/inputs/Address.tsx
+++ b/src/inputs/Address.tsx
@@ -39,7 +39,7 @@ class Address extends Component<IAddressProps> {
   }
 
   private get fieldSet () {
-    const { fieldConfig: { colProps, required, stateProps } } = this.injected
+    const { fieldConfig: { colProps, required, stateProps, disabled } } = this.injected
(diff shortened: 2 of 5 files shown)


rkuykendall:
So reading this, I thought "wait, doesn't fieldConfig have a disabled attribute?" and yes, yes it does. But it's... not implemented? It looks like it's a holdover from V2 where it IS implemented. I think you could add this line, change the test from `editProps: { disabled: true }` to `disabled: true` and be done?

image

OR, we should remove it from fieldConfig. But it seems important and universal enough to be included top-level. Agree we should implement it for everything else soon.

src/building-blocks/FormField.tsx
```diff
@@ -29,6 +29,7 @@ class FormField extends Component<IFormFieldProps> {
       ;
 
     return {
+      disabled: fieldConfig.disabled,
```
rpalermodrums: Why is this necessary? won't this be in `editProps`?

src/inputs/Address.tsx
```diff
@@ -39,7 +39,7 @@ class Address extends Component<IAddressProps> {
   }
 
   private get fieldSet () {
-    const { fieldConfig: { colProps, required, stateProps } } = this.injected
+    const { fieldConfig: { colProps, required, stateProps, disabled } } = this.injected
```
rpalermodrums: I thought that `Address` was outside of this use case?

src/building-blocks/FormField.tsx
rkuykendall: Na, it's a core attribute:
image

src/inputs/Address.tsx
rkuykendall: It was easy to implement it for all types.

PR #275, opened 2020-03-10
Add isCancelButtonDisabled; disabled docs page; improve types, tests

rkuykendall:
image

### DIFF
--- a/src/utilities/FormManager.ts
+++ b/src/utilities/FormManager.ts
@@ -1,3 +1,4 @@
+import { ComponentClass } from 'react';
 import { observable } from 'mobx';
 import autoBindMethods from 'class-autobind-decorator';
 import flattenObject from 'flat';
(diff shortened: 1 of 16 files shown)


src/utilities/FormManager.ts
```diff
@@ -80,20 +79,24 @@ class FormManager {
...
   }
 
-  public get submitButtonDisabled () {
+  public get isSubmitButtonDisabled (): boolean {
     return this.hasErrors();
   }
 
-  public getDefaultValue (fieldConfig: IFieldConfig) {
+  public get isCancelButtonDisabled (): boolean {
+    return this.saving;
```
rpalermodrums: in line with other changes, should be `isSaving`?

rkuykendall: Agreed.

PR #280, opened 2020-03-26
Remove FieldSet Divider

rkuykendall:
## Screenshots
**Mighty:**
image

**Before:**
image

**After:**
image

### DIFF
--- a/src/building-blocks/FieldSet.tsx
+++ b/src/building-blocks/FieldSet.tsx
@@ -8,6 +8,7 @@ import * as Antd from 'antd';
 import { IFieldSetPartial } from '../interfaces';
 import { IClassName } from '../props';
 import { isPartialFieldSetSimple } from '../utilities';
+import { CLASS_PREFIX } from '../consts';
 
 import Legend from './Legend';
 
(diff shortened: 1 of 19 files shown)


SHA-4:
Could you rename this PR to something more appropriate? `FieldSet` is already being used in `Form`.

Also, does removing the divider change anything visually? I don't see a difference in the before and after pictures

src/building-blocks/FieldSet.tsx
```diff
@@ -25,11 +25,13 @@ export default class FieldSet extends Component<IProps> {
       ;
 
     return (
-      <Antd.Row {...rowProps} className={cx(className)}>
-        <Legend fieldSet={fieldSet} />
+      <div className={cx('fields-ant-field-set', className)}>
```
SHA-4: I'm not sure what the purpose of a wrapper div would be here. Any reason in particular?

rkuykendall: Just styling, so you can add horizontal borders without messing with the styling / margins on Rows. This is what we use now, and from what I can tell it wouldn't be wise to do the bordering on the Row component.

rkuykendall: Could be removed if we find out otherwise after style changes. Doesn't make a difference though.

SHA-4: Sounds good

SHA-4: I would recommend having a const `CLASS_PREFIX = 'fields-ant'` or something similar for when we add classes to other components inevitability

PR #283, opened 2020-03-27
Fix type of className

rkuykendall:
**Bug:**
image

**Problem:** `export type IClassName = ClassValue[];` should be `ClassValue`

**Solution:** Just use `ClassValue` throughout library.


rkuykendall:
Because Travis and Github aren't on speaking terms right now:
image

PR #285, opened 2020-04-08
Hide table pagination by default on single-page tables

rkuykendall:
**Before:**
Screenshot 2020-04-08 09 34 34

**After:**
Screenshot 2020-04-08 09 33 09

### DIFF
--- a/src/components/Table.tsx
+++ b/src/components/Table.tsx
@@ -2,16 +2,19 @@ import React, { Component } from 'react';
 import { computed } from 'mobx';
 import { observer } from 'mobx-react';
 import autoBindMethods from 'class-autobind-decorator';
-import { omit } from 'lodash';
 import cx from 'classnames';
 
 import * as Antd from 'antd';
+import { TableProps } from 'antd/lib/table/interface';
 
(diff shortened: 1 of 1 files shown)


src/components/Table.tsx
```diff
@@ -26,27 +29,28 @@ class Table extends Component<ITableProps> {
...
 
     return (
       <Antd.Table
-        {...omit(this.props, 'title')}
+        {...passDownProps}
         className={cx(CLASS_NAME, className)}
         columns={this.columns}
         dataSource={this.dataSource}
         loading={isLoading}
+        pagination={{ hideOnSinglePage: true, ...this.props.pagination }}
```
rpalermodrums, line 53: Interesting that this is a prop. why all of the other changes besides this one?

rkuykendall:
@rpalermodrums I wanted to add that `hideOnSinglePage: true` to this table, and realized it actually doesn't extends or pass down the Ant Design table props. So this fixes two problems: 1) Defaults pagination to not show if only 1 page, and 2) Allow this component to take Antd.Table props in the future.

PR #289, opened 2020-07-23
Format backend validation response with array type

ekatsuta:
Format backend validation response with array type, in addition to objects

### DIFF
--- a/src/utilities/backendValidation.ts
+++ b/src/utilities/backendValidation.ts
@@ -44,8 +44,7 @@ function assignErrorFieldsToFormFields (
   // Try to assign error fields to form fields, falling back on generic array
   Object.keys(fieldErrors).forEach(errorField => {
     const message = fieldErrors[errorField]
-      , label = errorField === 'non_field_errors' ? '' : varToLabel(errorField)
-      ;
+      , label = errorField === 'non_field_errors' ? '' : varToLabel(errorField);
 
     // Check for an exact match
     if (fieldNames.includes(errorField)) {
--- a/src/utilities/backendValidation.ts
+++ b/src/utilities/backendValidation.ts
@@ -4,7 +4,6 @@ import { mapBooleanToText, varToLabel } from '@mighty-justice/utils';
 
 import { IBackendValidation, IErrorMessage, IFoundOnForm, toastError } from './FormManager';
 
-
 // Takes an API response and converts it to a string to string map
 function getFieldErrors (errors: { [key: string]: any }, prefix = '') {
   const messages: { [key: string]: string } = {};
--- a/src/utilities/backendValidation.ts
+++ b/src/utilities/backendValidation.ts
@@ -2,7 +2,8 @@ import { isArray, isPlainObject, isString, extend, isBoolean } from 'lodash';
 
 import { mapBooleanToText, varToLabel } from '@mighty-justice/utils';
 
-import { IBackendValidation, IErrorMessage, IFoundOnForm } from './FormManager';
+import { IBackendValidation, IErrorMessage, IFoundOnForm, toastError } from './FormManager';
+
 
 // Takes an API response and converts it to a string to string map
 function getFieldErrors (errors: { [key: string]: any }, prefix = '') {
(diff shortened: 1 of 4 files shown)


test/utilities/backendValidation.test.tsx
```diff
@@ -59,6 +59,11 @@ const TESTING_PAIRS = [
     FOUND_ON_FORM: {},
     ERROR_MESSAGES: [{ field: 'Is Account Created', message: 'No' }],
   },
+  {
+    RESPONSE: [getString(6)],
```
rkuykendall, line 63: Could you add a test for
```ts
    RESPONSE: getString(6),
```
if it passes without much work?

src/utilities/backendValidation.ts
```diff
@@ -74,3 +75,29 @@ export default function backendValidation (fieldNames: string[], response: objec
...
+  }
+
+  if (isString(response)) {
+    return {
+      errorMessages: [{ field: '', message: response }],
+      foundOnForm: {},
+    };
+  }
+
+  return {
```
rkuykendall, line 99: @ekatsuta Hm this is uncovered, maybe we should add a `null` test?

src/utilities/backendValidation.ts
```diff
@@ -33,9 +33,10 @@ function getFieldErrors (errors: { [key: string]: any }, prefix = '') {
   return messages;
 }
 
-export default function backendValidation (fieldNames: string[], response: object): IBackendValidation {
-  const fieldErrors = getFieldErrors(response)
-    , foundOnForm: IFoundOnForm = {}
+function assignErrorFieldsToFormFields (
+    fieldNames: string[], fieldErrors: { [key: string]: string },
+  ): IBackendValidation {
```
SHA-4, line 38: [nit] splitting the function definition but not splitting the arguments

src/utilities/backendValidation.ts
SHA-4: Should `null` have a default message? As it stands now, wouldn't this swallow up the error?

src/utilities/backendValidation.ts
```diff
@@ -74,3 +75,29 @@ export default function backendValidation (fieldNames: string[], response: objec
 
   return { errorMessages, foundOnForm };
 }
+
+export default function backendValidation (fieldNames: string[], response: any): IBackendValidation {
+  if (isArray(response)) {
+    return {
+      errorMessages: [{ field: '', message: response[0] }],
```
SHA-4, line 82: Do you think it's better just to display the first? Or should the errors be joined in some way?

ekatsuta: I used the first b/c that's how we're determining the `fieldErrors` in `getFieldErrors` method...thoughts? otherwise, might make sense to join with `;` ?

rkuykendall: We could map them, like `response.map(message => ({ field: '', message }))`

This gets tricky honestly because what if the backend returned an array of objects? We could handle a lot more use cases here, which I don't love. But for now, I say keep it to things we've seen "in the field"

src/utilities/backendValidation.ts
rkuykendall: If we return `errorMessages: [{ field: '', message: '' }]` it will display the default.

test/utilities/backendValidation.test.tsx
```diff
@@ -59,6 +59,21 @@ const TESTING_PAIRS = [
...
+  },
+  {
+    RESPONSE: getString(7),
+    FOUND_ON_FORM: {},
+    ERROR_MESSAGES: [{ field: '', message: getString(7) }],
+  },
+  {
+    RESPONSE: null,
+    FOUND_ON_FORM: {},
+    ERROR_MESSAGES: [{ field: '', message: 'Response was null.' }],
```
rkuykendall, line 75: This is not a very user-friendly message. I'd go with the const `toastError` which is currently "Error submitting form" but can be improved to be even more friendly in the future!

rkuykendall: Like "There was an error submitting the form." would be neat and profesional 😮

PR #290, opened 2020-08-11
Fix: FieldSet changes shown on page but not included in submit model

rkuykendall:
This bug was discovered during Emi's work on the dynamic cases form. A test existed for this use-case, but the test was insufficient. This PR improves the test suite and fixes the bug.

### DIFF
--- a/src/utilities/FormManager.ts
+++ b/src/utilities/FormManager.ts
@@ -23,6 +23,7 @@ import { IModel, IValue } from '../props';
 
 import backendValidation from './backendValidation';
 import { getFieldSetsFields, modelFromFieldConfigs } from './common';
+import { fillInFieldSets } from './fillIn';
 
 export interface IFoundOnForm { [key: string]: string; }
 export interface IErrorMessage { field: string; message: string; }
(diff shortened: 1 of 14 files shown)


src/utilities/FormManager.ts
```diff
@@ -84,15 +85,21 @@ class FormManager {
...
-
   public get isFormDisabled (): boolean {
     // The disabled prop can be changed any time, so we can't just save it locally
     return this.isSaving || !!this.formWrappedInstance.props.disabled;
   }
 
+  public get fieldSets (): IFieldSet[] {
+    // The fieldSets prop can be changed any time, so try to get them dynamically if you can
+    const fieldSetsProp = this.formWrappedInstance.props.fieldSets;
+    return fieldSetsProp ? fillInFieldSets(fieldSetsProp) : this.args.fieldSets;
```
SHA-4, line 96: Is there any case when we would want to defer to the passed field sets instead of the form props? I can't think of one, but just want to make sure

test/features/respondToFieldsetsChange.test.ts
```diff
@@ -11,27 +11,75 @@ const SUPPORTING_COMPONENTS = [
...
-
-      const tester = await new Tester(ComponentClass, { props }).mount({async: true});
+    it(`${componentName} successfully updates when field name changes`, async () => {
+      const { ComponentClass } = COMPONENT_GENERATORS[componentName]
+        , props = getProps(componentName)
+        , oldProps = {...props, fieldSets: [[{ field: 'originalName', type: 'string' }]]}
+        , newProps = {...props, fieldSets: [[{ field: 'updatedName', type: 'string' }]]}
+        ;
+
+      const tester = await new Tester(ComponentClass, { props: oldProps }).mount({async: true});
```
SHA-4, line 38: Tester mounts async by default, unless this is an older version of tester (it doesn't look like it is)

test/features/respondToFieldsetsChange.test.ts
```diff
@@ -11,27 +11,75 @@ const SUPPORTING_COMPONENTS = [
...
+      await changeProps(tester, newProps);
+
+      expect(tester.find('input[id="originalName"]').length).toBe(0);
       expect(tester.find('input[id="updatedName"]').length).toBe(1);
+
+      tester.submit();
+      expect(props.onSave).toHaveBeenCalledWith({ updatedName: '' });
+    });
+
+    it(`${componentName} successfully updates when number of fieldsets change`, async () => {
```
SHA-4, line 57: Since field sets is camel-cased throughout the app, I would expect it to be `field sets`, but I recognize that that seems counterintuitive

src/utilities/FormManager.ts
rkuykendall: No, I don't think so. The reason we're doing this is in the off chance that `formWrappedInstance` doesn't take a `fieldSets` prop, like if someone instantiates it themselves somewhere.

test/features/respondToFieldsetsChange.test.ts
rkuykendall: I'll make it FieldSets, since that's the name of the interface for this type of object.

PR #354, opened 2021-02-24
Add left-right label-data support

wsaffran:
Restructure fields-ant to support left-right label-data display
- Card objects can display label-data vertically or inline with the optional `display` prop [`"horizontal"`, `"vertical"`, or `"inline"`]
- Card objects can display a `:` postfixed to the label with the optional `colon` prop [`boolean`]
- Form objects can display label-data vertically or inline with the optional `display` prop [`"horizontal"`, `"vertical"`, or `"inline"`]
- Form objects can display a `:` postfixed to the label with the optional `colon` prop [`boolean`]
- Defaults for Card is `"vertical"` and Form label-data display is `"horizontal"` (this is currently how it is on our app)
- `"horizontal"` is equivalent to the default non-form Card objects

### Screenshots:

Card left/right:
Screen Shot 2021-02-24 at 4 55 26 PM
Screen Shot 2021-02-24 at 4 55 16 PM

Card vertical:
Screen Shot 2021-02-24 at 4 53 34 PM
Screen Shot 2021-02-24 at 4 53 22 PM

Form left/right:
Screen Shot 2021-02-24 at 4 46 16 PM

Form vertical:
Screen Shot 2021-02-24 at 4 45 34 PM

Colon options:
Screen Shot 2021-02-26 at 9 37 47 AM
Screen Shot 2021-02-26 at 9 37 59 AM
Screen Shot 2021-02-25 at 9 02 46 AM

### DIFF
--- a/src/building-blocks/Info.tsx
+++ b/src/building-blocks/Info.tsx
@@ -16,7 +16,7 @@ class Info extends Component<{ fieldConfig: IFieldConfigPartial; format?: IForma
   public render() {
     const { format } = this.props,
       layout = format && format.layout,
-      rowClassName = layout !== 'inline' ? '' : `${CLASS_PREFIX}-info-item-inline`;
+      rowClassName = `${CLASS_PREFIX}-info-row-${layout}`;
 
     return (
       <Antd.Col {...this.props.fieldConfig.colProps} className={`${CLASS_PREFIX}-info`}>
--- a/src/building-blocks/Info.tsx
+++ b/src/building-blocks/Info.tsx
@@ -8,7 +8,8 @@ import { ClassValue } from 'classnames/types';
 import * as Antd from 'antd';
 
 import { CLASS_PREFIX } from '../consts';
-import { IFieldConfigPartial } from '../interfaces';
+import { IFieldConfigPartial, IFormatProps } from '../interfaces';
+import { getLabelAfter } from '../utilities/common';
 
 @autoBindMethods
 @observer
@@ -24,17 +25,38 @@ class Info extends Component<{ fieldConfig: IFieldConfigPartial }> {
 
 @autoBindMethods
 @observer
-class Label extends Component<{ className?: ClassValue }> {
+class Label extends Component<{ className?: ClassValue, format?: IFormatProps }> {
   public render() {
-    return <div className={cx(this.props.className, `${CLASS_PREFIX}-info-label`)}>{this.props.children}</div>;
+    const { className, format } = this.props
+      , colon = format && format.colon
+      , layout = format && format.layout
+      , colonText = colon === false ? '' : ':'
+      , labelClassName = cx(className, `${CLASS_PREFIX}-info-label`)
+      , labelAfterClassName = cx(className, `${CLASS_PREFIX}-info-label-after`);
+
+    return layout === "inline"
+      ? <div style={{"float": "left"}} className={labelClassName}>
+          {this.props.children}
+          {getLabelAfter(labelAfterClassName, colonText)}
+        </div>
+      : <div className={labelClassName}>
+          {this.props.children}
+          {getLabelAfter(labelAfterClassName, colonText)}
+        </div>;
   }
 }
 
 @autoBindMethods
 @observer
-class Value extends Component<{ className?: ClassValue }> {
+class Value extends Component<{ className?: ClassValue, format?: IFormatProps }> {
   public render() {
-    return <div className={cx(this.props.className, `${CLASS_PREFIX}-info-value`)}>{this.props.children}</div>;
+    const { format } = this.props
+      , layout = format && format.layout
+      , className = cx(this.props.className, `${CLASS_PREFIX}-info-value`);
+
+    return layout === "inline"
+      ? <div style={{"float": "left"}} className={className}>{this.props.children}</div>
+      : <div className={className}>{this.props.children}</div>;
   }
 }
 
--- a/src/building-blocks/Info.tsx
+++ b/src/building-blocks/Info.tsx
@@ -40,7 +40,7 @@ class Label extends Component<{ className?: ClassValue; format?: IFormatProps }>
         `${CLASS_PREFIX}-info-label`,
         `${CLASS_PREFIX}-info-label${hasColon ? '' : '-no'}-colon`,
         `${CLASS_PREFIX}-info-label-layout-${layout}`,
-      )
+      );
 
     return (
       <div className={labelClassName}>
--- a/src/building-blocks/Info.tsx
+++ b/src/building-blocks/Info.tsx
@@ -25,38 +25,44 @@ class Info extends Component<{ fieldConfig: IFieldConfigPartial }> {
 
 @autoBindMethods
 @observer
-class Label extends Component<{ className?: ClassValue, format?: IFormatProps }> {
+class Label extends Component<{ className?: ClassValue; format?: IFormatProps }> {
   public render() {
-    const { className, format } = this.props
-      , colon = format && format.colon
-      , layout = format && format.layout
-      , colonText = colon === false ? '' : ':'
-      , labelClassName = cx(className, `${CLASS_PREFIX}-info-label`)
-      , labelAfterClassName = cx(className, `${CLASS_PREFIX}-info-label-after`);
+    const { className, format } = this.props,
+      colon = format && format.colon,
+      layout = format && format.layout,
+      colonText = colon === false ? '' : ':',
+      labelClassName = cx(className, `${CLASS_PREFIX}-info-label`),
+      labelAfterClassName = cx(className, `${CLASS_PREFIX}-info-label-after`);
 
-    return layout !== "inline"
-      ? <div className={labelClassName}>
-          {this.props.children}
-          {getLabelAfter(labelAfterClassName, colonText)}
-        </div>
-      : <div style={{"float": "left"}} className={labelClassName}>
-          {this.props.children}
-          {getLabelAfter(labelAfterClassName, colonText)}
-        </div>;
+    return layout !== 'inline' ? (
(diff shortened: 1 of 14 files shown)


src/building-blocks/Info.tsx
```diff
@@ -8,25 +8,41 @@ import { ClassValue } from 'classnames/types';
...
 @autoBindMethods
 @observer
-class Label extends Component<{ className?: ClassValue }> {
+class Label extends Component<{ className?: ClassValue; format?: IFormatProps }> {
   public render() {
-    return <div className={cx(this.props.className, `${CLASS_PREFIX}-info-label`)}>{this.props.children}</div>;
+    const { className, format } = this.props,
+      colon = format && format.colon,
+      layout = format && format.layout,
+      // colon can be displayed only when layout is inline or horizontal
```
rkuykendall, line 36: Wait, why?

wsaffran: This was how ant design's colon feature worked so I just tried to keep it consistent with that. My rationale was that on a FormCard, if there is a series of vertically aligned label-data fields with a colon, when someone switched to the edit mode all the colons would disappear because now it would be an ant design form which doesn't allow colons in that orientation

src/building-blocks/Info.tsx
```diff
@@ -8,25 +8,41 @@ import { ClassValue } from 'classnames/types';
...
+import { IFieldConfigPartial, IFormatProps } from '../interfaces';
 
 @autoBindMethods
 @observer
-class Info extends Component<{ fieldConfig: IFieldConfigPartial }> {
+class Info extends Component<{ fieldConfig: IFieldConfigPartial; format?: IFormatProps }> {
   public render() {
+    const { format } = this.props,
+      layout = format && format.layout,
+      rowClassName = layout !== 'inline' ? '' : `${CLASS_PREFIX}-info-item-inline`;
```
rkuykendall, line 19: Why only include a class sometimes?
```suggestion
      rowClassName = `${CLASS_PREFIX}-info-row-${layout}`;
```

(actually asking)

src/building-blocks/Info.tsx
```diff
@@ -8,25 +8,41 @@ import { ClassValue } from 'classnames/types';
...
-class Label extends Component<{ className?: ClassValue }> {
+class Label extends Component<{ className?: ClassValue; format?: IFormatProps }> {
   public render() {
-    return <div className={cx(this.props.className, `${CLASS_PREFIX}-info-label`)}>{this.props.children}</div>;
+    const { className, format } = this.props,
+      colon = format && format.colon,
+      layout = format && format.layout,
+      // colon can be displayed only when layout is inline or horizontal
+      colonClassName =
+        colon === false || layout === 'vertical' ? `${CLASS_PREFIX}-info-label-no-colon` : `${CLASS_PREFIX}-info-label`,
```
rkuykendall, line 38: Since  `${CLASS_PREFIX}-info-label` is used so much, I wonder if we should keep it, and add a class for colon and no-colon. And hell, we could add layout as well.

`cx(
  `${CLASS_PREFIX}-info-label`,
  `${CLASS_PREFIX}-info-label${colon ? '' : '-no'}-colon`
  `${CLASS_PREFIX}-info-label-layout-${layout}`
)`

src/building-blocks/Info.tsx
wsaffran: I guess this was because I didn't need any styles for the non-inline version of the row, but that makes sense that there should probably be a class name regardless

src/components/Form.tsx
```diff
@@ -91,13 +91,18 @@ export class UnwrappedForm extends Component<IFormWrappedProps> {
...
-    const { showControls, title } = this.props,
+    const { showControls, title, layout, colon } = this.props,
       formModel = this.formManager.formModel,
       filteredFieldSets = filterFieldSets(this.fieldSets, { model: formModel }),
       className = cx(CLASS_NAME, this.props.className);
 
     return (
-      <Antd.Form layout="vertical" onSubmit={this.formManager.onSave} className={className}>
+      <Antd.Form
+        layout={layout || 'horizontal'}
```
rkuykendall, line 101: I think this could be fixed by setting a default in `formPropsDefaults`, or since it's not a formProp, we could add a 

```ts
export const sharedComponentPropsDefaults = {
  layout: 'horizontal',
  colon: false,
};
```

and add it to these components like formPropsDefaults is today.

wsaffran: This sounds good, the only thing right now is that Card components default to vertical orientation and Form components default to horizontal (because of and design). Defaulting both to the same would in one way or another change some components on the app when this gets released. Wondering your thoughts on this?

wsaffran: Actually, that shouldn't be a problem, it looks like we were defaulting to vertical previously for our forms. Will just make "vertical" the default then and that should work for both!

src/building-blocks/Info.tsx
rkuykendall: This makes sense!

src/building-blocks/Info.tsx
rkuykendall: Yeah best to think of it as a general-purpose library, if only to save ourselves from having to do releases just to add a class when writing CSS elsewhere.

src/components/Form.tsx
rkuykendall: Perfect! Although, before, why were't cards getting your `sharedComponentPropsDefaults` default?

src/components/Form.tsx
```diff
@@ -91,13 +91,18 @@ export class UnwrappedForm extends Component<IFormWrappedProps> {
...
       filteredFieldSets = filterFieldSets(this.fieldSets, { model: formModel }),
       className = cx(CLASS_NAME, this.props.className);
 
     return (
-      <Antd.Form layout="vertical" onSubmit={this.formManager.onSave} className={className}>
+      <Antd.Form
+        layout={layout}
+        onSubmit={this.formManager.onSave}
+        className={className}
+        colon={colon === false ? false : true}
```
rkuykendall, line 104: With prop defaults, you can just do this, right?
```suggestion
        colon={colon}
```

src/components/Form.tsx
```diff
@@ -91,13 +91,18 @@ export class UnwrappedForm extends Component<IFormWrappedProps> {
...
-    const { showControls, title } = this.props,
+    const { showControls, title, layout, colon } = this.props,
       formModel = this.formManager.formModel,
       filteredFieldSets = filterFieldSets(this.fieldSets, { model: formModel }),
       className = cx(CLASS_NAME, this.props.className);
 
     return (
-      <Antd.Form layout="vertical" onSubmit={this.formManager.onSave} className={className}>
+      <Antd.Form
+        layout={layout}
```
rkuykendall, line 101: Sort these props

src/components/Form.tsx
rkuykendall: OH, because you are't using `sharedComponentPropsDefaults` in your display components because you are using `format` as a single prop and not `colon` and `layout` like you do in the other components. Why is that?

test/features/labelDataDisplay.test.ts
```diff
@@ -0,0 +1,81 @@
...
+}
+
+describe('Renders', () => {
+  COMPONENTS.forEach(componentName => {
+    LAYOUTS.forEach(layout => {
+      it(`Displays ${layout} for ${componentName}`, async () => {
+        const { ComponentClass, propsFactory } = COMPONENT_GENERATORS[componentName],
+          props = propsFactory.build({
+            fieldSets: [[normalFieldSet]],
+            layout: layout,
```
rkuykendall, line 29:
```suggestion
            layout,
```

test/features/labelDataDisplay.test.ts
```diff
@@ -0,0 +1,81 @@
+import { Tester } from '@mighty-justice/tester';
+import { COMPONENT_GENERATORS, fakeField, fakeTextShort } from '../factories';
+
+const normalFieldSet = { field: fakeField(), label: fakeTextShort(), type: 'string' },
```
rkuykendall, line 4: Looks like you could just use the factory:

```suggestion
const fieldSets = [[stringFactory.build()]],
```

And if you add the [[]] here and rename it `fieldSets` you can shorthand it below, like I showed with layout.

test/features/labelDataDisplay.test.ts
```diff
@@ -0,0 +1,81 @@
...
+      it(`Displays ${layout} when reading and writing for ${componentName}`, async () => {
+        const { ComponentClass, propsFactory } = COMPONENT_GENERATORS[componentName],
+          props = propsFactory.build({
+            fieldSets,
+            layout,
+          }),
+          tester = await new Tester(ComponentClass, { props }).mount();
+
+        expect(isForm(tester)).toBe(false);
+        expect(tester.find(`.fields-ant-info-row-${layout}`).length);
```
rkuykendall, line 70: Does this actually test anything without a `.toBe(1)` in it? Like does it default to checking if the value is truthy? Or does it need something added to the end?

wsaffran: Ah yeah that's a good catch. I assumed that expect would default to false for 0 which is was not the case. I updated the tests and removed the colon test for the `CARD_COMPONENTS` because fields ant handles the colon with an `::after`. Do you think it makes sense still to add a class like `fields-ant-form-label-${layout}` to the form and test it?

rkuykendall: Let's leave it for now and add it if we need it, unless you think it's good for uniformity and consistency.

wsaffran: Actually, I added a PR for this and to test component defaults so if it's overkill I can easily revert a commit if needed

rkuykendall (approved):

PR #364, opened 2021-03-03
Match layout features to ant-design's form

wsaffran:
Updated the `horizontal` and `inline` `layout` options to reflect that of ant design's [Form](https://ant.design/components/form/) `layout` options.

Patch for [MTY-3412](https://thatsmighty.atlassian.net/browse/MTY-3412)

### Screenshots:
All screenshots are of a form component with no row/col props and completely unformatted. Only differences are layout types

**Vertical [default]:**
Screen Shot 2021-03-04 at 9 56 55 AM

**Horizontal:**
Screen Shot 2021-03-04 at 9 56 21 AM

**Inline:**
Screen Shot 2021-03-04 at 9 56 43 AM

### DIFF
--- a/src/building-blocks/FormField.tsx
+++ b/src/building-blocks/FormField.tsx
@@ -2,7 +2,6 @@ import React, { Component } from 'react';
 import { computed } from 'mobx';
 import { observer } from 'mobx-react';
 import autoBindMethods from 'class-autobind-decorator';
-import { omit } from 'lodash';
 
 import { FormManager, fillInFieldConfig, filterFieldConfig } from '../utilities';
 import { IFieldConfigPartial } from '../interfaces';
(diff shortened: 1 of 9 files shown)


src/building-blocks/FormField.tsx
```diff
@@ -48,15 +49,20 @@ class FormField extends Component<IFormFieldProps> {
...
       return <EditComponent {...this.editProps} />;
     }
 
     return (
-      <FormItem fieldConfig={this.fieldConfig} formManager={formManager} formModel={formModel}>
+      <FormItem
+        fieldConfig={this.fieldConfig}
+        formManager={formManager}
+        formModel={formModel}
+        {...omit(passDownProps, ['fieldConfig'])}
```
rkuykendall, line 64: if you move `{...passDownProps}` to the top of this list, `fieldConfig={` will override it. I can't decide if that's better because we don't use omit or worse because it breaks on sort and is less verbose. Your call. Maybe put fieldconfig at the bottom with a note of we decide to go that way, or just keep it like this.

rkuykendall (approved):

PR #366, opened 2021-03-05
Add form class names for colon

wsaffran:
Add `className` for colon on `formItem`. This is a requirement to add styles in registry.

Patch for [MTY-3412](https://thatsmighty.atlassian.net/browse/MTY-3412)

### DIFF
--- a/src/building-blocks/FormItem.tsx
+++ b/src/building-blocks/FormItem.tsx
@@ -7,7 +7,7 @@ import cx from 'classnames';
 import * as Antd from 'antd';
 import { ValidationRule as AntValidationRule } from 'antd/es/form';
 
-import { FormManager, noopValidator, renderLabel } from '../utilities';
+import { formatClassNames, FormManager, noopValidator, renderLabel } from '../utilities';
 import { IFieldConfig, IFieldsValidator, ILayout } from '../interfaces';
 import { IModel } from '../props';
 import { CLASS_PREFIX } from '../consts';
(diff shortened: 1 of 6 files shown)


src/components/Form.tsx
```diff
@@ -96,7 +96,7 @@ export class UnwrappedForm extends Component<IFormWrappedProps> {
       filteredFieldSets = filterFieldSets(this.fieldSets, { model: formModel }),
       hasColon = !(colon === false || layout === 'vertical'),
       className = cx(CLASS_NAME, this.props.className, `${CLASS_NAME}${hasColon ? '' : '-no'}-colon`),
-      passDownProps = { layout: layout };
+      passDownProps = { layout: layout, colon: colon };
```
rkuykendall, line 99:
```suggestion
      passDownProps = { layout, colon };
```

src/building-blocks/FormItem.tsx
```diff
@@ -97,13 +98,15 @@ class FormItem extends Component<IFormFieldProps> {
...
-    const { formManager, fieldConfig, layout } = this.props,
+    const { formManager, fieldConfig, layout, colon } = this.props,
       { colProps, formItemProps, field } = fieldConfig,
+      hasColon = !(colon === false || layout === 'vertical'),
       className = cx(
         FORM_ITEM_CLASS_NAME,
         fieldConfig.className,
         formItemProps && formItemProps.className,
         `${FORM_ITEM_CLASS_NAME}-${layout}`,
+        `${FORM_ITEM_CLASS_NAME}${hasColon ? '' : '-no'}-colon`,
```
rkuykendall, line 109: I think this bit of code is in 3 places now? We should probably group it into a util, something like:

```ts
function formatClassNames (className: string, layout, colon: boolean) {
  const hasColon = !(colon === false || layout === LAYOUT_TYPES.VERTICAL);
  return {
    `${FORM_ITEM_CLASS_NAME}-${layout}`,
    `${FORM_ITEM_CLASS_NAME}${hasColon ? '' : '-no'}-colon`,
  }
}
```
then `{...formatClassNames(FORM_ITEM_CLASS_NAME, layout, colon)}`

rkuykendall: Also we should have a const for the layout strings.

rkuykendall (approved):

PR #368, opened 2021-03-09
Fix: FormCard class names

wsaffran:
Make sure `FormCard` uses the `className` prop


rkuykendall:
This seems like an excellent case for a test. Perhaps we should move your new `EDITABLE_COMPONENTS` and `CARD_COMPONENTS` const into factories and use the latter to test they all accept and use className props?

rkuykendall (approved):

PR #408, opened 2021-08-05
Make event optional for onSave in FormManager

ellenshin:
Mainly for this registry PR: https://github.com/mighty-justice/registry-client/pull/2695

### DIFF
--- a/src/utilities/FormManager.ts
+++ b/src/utilities/FormManager.ts
@@ -286,8 +286,10 @@ class FormManager {
     }
   }
 
-  public async onSave(event: any) {
-    event.preventDefault();
+  public async onSave(event?: any) {
+    if (event) {
+      event.preventDefault();
+    }
 
     this.isSaving = true;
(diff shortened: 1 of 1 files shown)


src/utilities/FormManager.ts
```diff
@@ -286,8 +286,10 @@ class FormManager {
     }
   }
 
-  public async onSave(event: any) {
-    event.preventDefault();
+  public async onSave(event?: any) {
+    if (!!event) {
```
rkuykendall, line 290: Hm, technically, the `!!` doesn't affect the if statement, but I'm also fine with it for readability, your call.

```suggestion
    if (event) {
```

rkuykendall (approved):

src/utilities/FormManager.ts
ellenshin: no! ill take that suggestion. it's simplier

PR #435, opened 2021-11-12
Make argument for formManager's onFinish optional

wsaffran:
Make argument for `formManager`'s `onFinish` function optional. This causes errors in the Registry where `onFinish` is called without any arguments

### DIFF
--- a/src/utilities/FormManager.ts
+++ b/src/utilities/FormManager.ts
@@ -267,7 +267,7 @@ class FormManager {
     this.notifyUserAboutErrors(backendErrors.errorMessages);
   }
 
-  public async onFinish(_values: any) {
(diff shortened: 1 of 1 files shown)


src/utilities/FormManager.ts
```diff
@@ -267,7 +267,7 @@ class FormManager {
     this.notifyUserAboutErrors(backendErrors.errorMessages);
   }
 
-  public async onFinish(_values: any) {
+  public async onFinish(_values?: any) {
```
rkuykendall, line 270: Does this work? Actually asking.

```suggestion
  public async onFinish() {
```

rkuykendall (approved):

PR #450, opened 2021-11-22
Fix: Date validation test act errors

wsaffran:
This is an attempt to fix the `jest.setTimeout.Error` on the `date.test.ts`'s `validate` test when trying to deploy a new version

Screen Shot 2021-11-22 at 11 25 04 AM

This is the first of two ideas to fix the issue:
1) Fix act errors
2) Add a default timeout for that particular test to give a greater buffer (maybe 10000ms)

**Note:** only the contents of [this pr](https://github.com/mighty-justice/fields-ant/pull/449) have been changed since the last version `v1.8.2`

### DIFF
--- a/test/types/date.test.ts
+++ b/test/types/date.test.ts
@@ -14,9 +14,12 @@ async function checkInferYear(newYear: string) {
     onSave = jest.fn(),
     props = { fieldSets, model, onSave };
 
-  const tester = await new Tester(FormCard, { props }).mount();
-  await tester.changeInput(`input[id="${fieldConfig.field}.year"]`, newYear);
-  await tester.submit();
+  const tester = new Tester(FormCard, { props });
+  await act(async () => {
+    await tester.mount();
+    await tester.changeInput(`input[id="${fieldConfig.field}.year"]`, newYear);
+    await tester.submit();
+  });
   return onSave.mock.calls[0][0][fieldConfig.field].slice(0, 4);
 }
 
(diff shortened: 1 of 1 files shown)


rpalermodrums (approved):
Highly suggest at least making the one change, but going to ✅ so that i'm not a blocker

test/types/date.test.ts
```diff
@@ -25,7 +27,10 @@ async function isInputsValid(month: string, day: string, year: string) {
...
     props = { fieldSets, onSave };
 
-  const tester = await new Tester(FormCard, { props }).mount();
+  const tester = new Tester(FormCard, { props });
+  await act(async () => {
+    await tester.mount();
+  });
   await tester.changeInput(`input[id="${fieldConfig.field}.month"]`, month);
   await tester.changeInput(`input[id="${fieldConfig.field}.day"]`, day);
   await tester.changeInput(`input[id="${fieldConfig.field}.year"]`, year);
```
rpalermodrums, line 36: Basically anything that isn't a precondition (like defining variables/factory instances or `onBeforeMount` setup) or assertion, and especially anything that interacts with the virtual DOM (even more especially for async interactions), should be wrapped in `act`.

So, all of the input interaction should also be wrapped, even though i'm pretty sure those awaits are doing nothing. I'd suggest that you fix the other tests in this file as well to future-proof, but all good if this is urgent. wrapping lines  34-38 would be sufficient

rpalermodrums: relevant docs:
- https://reactjs.org/docs/testing-recipes.html#act
- http://wiki.c2.com/?ArrangeActAssert

wsaffran: @rpalermodrums Yeah that makes sense. I left the rest of the actions as is because Robert updated our tester package to wrap everything that need to be wrapped in an `act` block like [`changeInput`](https://github.com/mighty-justice/tester/blob/master/src/tester.tsx#L114). though, it hasn't worked perfectly with `mount` and nested `act` blocks are okay so I updated the file to include all the items just in case

PR #451, opened 2021-11-29
Fix: Nested form item values

wsaffran:
Fix handling of nested for item values

### DIFF
--- a/src/utilities/common.tsx
+++ b/src/utilities/common.tsx
@@ -1,6 +1,5 @@
 import React from 'react';
 import cx from 'classnames';
-import flattenObject from 'flat';
 import { flatten as flattenArray, get, has, isArray, isObject, isString, kebabCase, set, some, sortBy } from 'lodash';
 
 import { ColumnProps } from 'antd/es/table';
@@ -26,6 +25,7 @@ import WithTooltip from '../building-blocks/WithTooltip';
 
 import { fillInFieldConfig, fillInFieldSets } from './fillIn';
 import { filterFieldConfig } from './filters';
+import { flatten } from './flatten';
 
 export function isPartialFieldSetSimple(fieldSet: IFieldSetPartial): fieldSet is IFieldSetSimplePartial {
   return isArray(fieldSet);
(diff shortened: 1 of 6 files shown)


src/utilities/common.tsx
```diff
@@ -203,3 +204,26 @@ export function formatClassNames(className: string, colon: boolean = false, layo
...
+        (accumulator[current] = Number.isNaN(Number(keys[index + 1]))
+          ? keys.length - 1 === index
+            ? flattenedObject[key]
+            : {}
+          : []),
+      result,
+    );
+  });
+
+  return result;
```
rkuykendall, line 228: This is a truly impressive piece of code. However, I think we can remove most of this logic by using `_.set` to handle the array and dot notation to object placement for us:

```ts
  const flattenedObject = flatten(object, { removeDate: true });

  return Object.entries(flattenedObject).reduce(
    (objOut, [key, value]) => set(objOut, key, value), {}
 );
```

src/utilities/flatten.ts
```diff
@@ -0,0 +1,45 @@
...
+
+const isBuffer = (obj: any) => {
+  return obj && obj.constructor && typeof obj.constructor.isBuffer === 'function' && obj.constructor.isBuffer(obj);
+};
+
+const keyIdentity = (key: string) => {
+  return key;
+};
+
+export const flatten = (target: Object, opts: any = {}) => {
```
rkuykendall, line 11: Is this from somewhere? And why this instead of `import flattenObject from 'flat';` like in FormManager?

wsaffran: Yep this is mostly taken from `flat`'s `flattenObject`. I added an `opts` to handle `Moment` objects (see `isDate` below) because the `flattenObject` was flattening the whole `Moment` object when we just want to keep it as a value, not a series of new key: value pairs. That was messing up all the `datePicker` inputs

src/utilities/common.tsx
wsaffran: Thanks! but I had some help from good 'ol stackOverflow with this one. And yeah I always forget about lodash will update

src/utilities/flatten.ts
rkuykendall: Ah interesting!

1. Can you add a comment to the code about all this?

Future ideas:
2. We should probably just remove datepicker soon, they are awful. We only use 1 in registry and it would be better as date
3. We could replace this with date-fns and reduce bundle size: https://ant.design/docs/react/replace-moment#Use-date-fns

wsaffran: Yeah that makes sense. Would be a good follow-up after ant-4 is fully released

rkuykendall (approved):

PR #490, opened 2022-02-07
Improve required error message

rkuykendall:
**Jira:** https://thatsmighty.atlassian.net/browse/PROD-664

Examples:

-  `Required - Please input a valid value` => `This field is required`
- `Required - Please input a valid Law Firm` => `'Law Firm' is required`
- `Required - Please input a valid Enter an external system identifier` => `'Enter an external system identifier' is required`

### DIFF
--- a/src/utilities/fillIn.ts
+++ b/src/utilities/fillIn.ts
@@ -89,13 +89,13 @@ function inferType(fieldConfig: Partial<IFieldConfig>) {
 }
 
 export function fillInFieldConfig(fieldConfig: IFieldConfigPartial): IFieldConfig {
-  const type = inferType(fieldConfig),
-    label = fieldConfig.label || varToLabel(getFieldSuffix(fieldConfig.field));
+  const type: string = inferType(fieldConfig),
+    label: string = fieldConfig.label || varToLabel(getFieldSuffix(fieldConfig.field));
 
   const requiredValidationRule: undefined | { [key: string]: Rule } = fieldConfig.required
     ? {
         required: {
-          message: ` ${`'${label}'` || 'This field'} is required`,
+          message: `'${label}' is required`,
           required: true,
         },
       }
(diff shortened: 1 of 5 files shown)


src/utilities/fillIn.ts
```diff
@@ -95,7 +95,7 @@ export function fillInFieldConfig(fieldConfig: IFieldConfigPartial): IFieldConfi
   const requiredValidationRule: undefined | { [key: string]: Rule } = fieldConfig.required
     ? {
         required: {
-          message: `Required - Please input a valid ${label || 'value'}`,
+          message: ` ${`'${label}'` || 'This field'} is required`,
```
tskela, line 98: I tried running this code with `label` as `undefined`, and the message printed is `'undefined' is required`. I'm assuming adding the quotation marks around the value of `label` causes the value to no longer be falsy.

rkuykendall: YUP, turns out label will always be defined, so I rethought the expected result and wrote tests.

test/utilities/fillIn.test.ts
```diff
@@ -0,0 +1,28 @@
...
+    const required = true,
+      field = 'field';
+
+    expect(getRequiredMessage({ required, field })).toBe(`'Field' is required`);
+
+    expect(getRequiredMessage({ required, field, label: undefined })).toBe(`'Field' is required`);
+
+    expect(getRequiredMessage({ required, field, label: 'Law Firm' })).toBe(`'Law Firm' is required`);
+
+    expect(getRequiredMessage({ required, field, label: 'Law Firm', showLabel: false })).toBe(`'Law Firm' is required`);
```
tskela, line 26: How should `showLabel` affect the resulting message? I don't see it being used.

rkuykendall: It's not. I guess this is just a check against future stuff that might mess with things. Or to at least acknowledge this use case.

PR #514, opened 2022-04-25
Add collapsible prop to editable array card

tskela:
The Jira ticket linked below states: "Ideally, past evaluations would be collapsed by default but expandable."

This PR modifies the `EditableArrayCard` component to allow for a `collapsible` prop. This determines whether the editable cards are rendered as usual, or as collapsible items within a `Collapse` component:

Screen Shot 2022-04-25 at 5 30 06 PM

By default, only the first item is shown and the others are collapsed (which means only the panel headers are rendered until the item is expanded).

Jira: https://thatsmighty.atlassian.net/browse/PROD-878?atlOrigin=eyJpIjoiZWY5NWI4ZTg5NTgxNDI2ODgxMTUwNGNjOTc5MWYzOGEiLCJwIjoiaiJ9

### DIFF
--- a/src/components/EditableArrayCard.tsx
+++ b/src/components/EditableArrayCard.tsx
@@ -5,7 +5,7 @@ import { isEmpty } from 'lodash';
 import autoBindMethods from 'class-autobind-decorator';
 import SmartBool from '@mighty-justice/smart-bool';
 
-import { Card } from 'antd';
+import { Card, Collapse } from 'antd';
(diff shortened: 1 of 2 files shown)


tskela:
Note: I wasn't sure if it was worth it to modify a `fields-ant` component for this one use case, but, when I tried to implement this as a component in CoOp, I was obviously repeating a lot of the `EditableArrayCard` code. Let me know if you think it's best to keep this in CoOp.

PR #574, opened 2023-08-24
Conditional btn render

ellenshin:


src/components/EditableCard.tsx
```diff
@@ -55,50 +64,66 @@ class EditableCard extends Component<IEditableCardProps> {
...
+        </span>
+      </Tooltip>
     );
   }
 
   private buttons () {
     return (
       <ButtonToolbar noSpacing>
         {this.deleteButton}
+        <div style={{ display: 'inline-block', width: '10px' }}></div>
```
cry-stal-lee, line 126: Can we try using an `antd` component for this? i.e.

import { Space, Button } from 'antd';

function MyComponent() {
  return (
    <Space size={10}>
      {this.deleteButton}
      {this.editButton}
    </Space>
  );
}

cry-stal-lee, line 126: Can we try using an `antd` `Space` for this? i.e.
```
import { Space } from 'antd';

function MyComponent() {
  return (
    <Space size={10}>
      {this.deleteButton}
      {this.editButton}
    </Space>
  );
}
```

ellenshin: `Space` component isnt supported in antd3

cry-stal-lee: hmm could we try making a parent `div` with `display: flex`, `gap: 10px`, and `justify-content: space-between` and see if that works? Would like to try to avoid the empty `<div>` if possible, though if that doesn't work I'm OK with this.

it also looks like we don't have much custom styling in this library since we mostly just use the built-in `antd` functions, so I think inline styling is OK as a workaround for now since this is just a branch. ideally on the `main` branch we will eventually upgrade to `antd4` and use `<Space>` for things like this.

ellenshin: actually deleting the `span`s made this not needed. im seeing the tooltips when the buttons are disabled so not sure what you saw in the previous PR? 🤔

ellenshin: maybe antd3 vs 4 difference?

cry-stal-lee (approved):

